PR #110, opened 2026-05-09
fix(missions): persist model selection per mission

emutis21:
## Who this is for

Anyone running several missions at the same time — a sales person researching three accounts in parallel, an operator drafting two emails while a long research task runs in another mission.

## The problem

Each mission has a model picker under the chat input — the little "Sonnet / Opus / Haiku" selector that controls which AI is doing the work.

Users want fast and cheap (Haiku) for short replies, and deeper but slower (Opus) for hard research. The picker LOOKED like it was per-mission, but it wasn't: switching it on one mission silently changed it on every other mission too.

Esteban Jaramillo flagged this in the beta-users WhatsApp group on Friday. Felipe confirmed it's a bug:

> "Es un bug. La manera en que debería funcionar es que cada vez que abras una misión, debajo de la barra de texto, puedes elegir el modelo."

## What's different now

The picker is honestly per-mission:

- Pick Opus on mission A → mission B is unaffected.
- Switch to Haiku on mission B → mission A still on Opus.
- Quit Houston and reopen → each mission remembers what you left it on.

If saving the choice ever fails (rare — would need disk to be unreachable), the user gets a visible toast instead of silent inconsistency, in line with the beta-stage error policy already documented in CLAUDE.md.

## Why it matters for the user

The picker is the only knob between "fast and cheap" and "deep but slow". Power users were either accidentally running their cheap work on Opus (burning tokens) or their hard work on Haiku (getting weaker answers) without realising. The bug had a direct cost-and-quality impact every time they switched contexts.

## How it was verified

- Repro confirmed on macOS desktop build: two missions side by side, model switched, restart, each mission held its choice.
- Engine unit test (`per_mission_model_isolation`) covers the isolation guarantee.
- App typecheck, locale validation (en/es/pt), and full engine test suite green.

Repro video below.

### DIFF
--- a/app/src/components/use-agent-chat-panel.tsx
+++ b/app/src/components/use-agent-chat-panel.tsx
@@ -36,13 +36,16 @@ import {
 } from "@houston-ai/chat";
 
 import { useFeedStore } from "../stores/feeds";
+import { useUIStore } from "../stores/ui";
 import { useWorkspaceStore } from "../stores/workspaces";
 import {
+  useActivity,
   useConnectedToolkits,
   useConnections,
   useSkills,
 } from "../hooks/queries";
 import {
+  tauriActivity,
   tauriAttachments,
   tauriChat,
   tauriConfig,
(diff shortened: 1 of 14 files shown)


emutis21:
https://github.com/user-attachments/assets/cab9280c-d321-4eee-b9a0-5d1b20b4daa1

PR #112, opened 2026-05-11
feat(chat): suggest first prompts when an agent has no skills yet

emutis21:
## Who this is for

Anyone opening Houston for the first time, plus power users who keep creating custom agents from scratch. They have an agent in front of them, an empty chat input, and no obvious next step.

## The problem

A Houston agent without skills shows an empty chat with no guidance. The user looks at "What should the agent work on?" and goes blank, or worse, types something the agent can't actually do because they don't know which integrations are connected.

This came up in the beta-users WhatsApp group on May 9, when one user asked how to actually use Apify inside Houston after connecting it. Another power user wrote a thirty-line starter prompt by hand and dropped it in the chat for the rest of the group. The product itself was not teaching this.

## What is different now

When the user opens a mission in an agent that has no skills, the empty area above the chat input now shows starter cards, in two flavors:

**If no integration is connected yet:**

A single "Connect a tool to get started" card. Clicking it opens the Integrations view.

**If at least one supported integration is connected (Gmail, LinkedIn, Apify, Notion, Sheets, Outlook, CRMs, call recorders, etc.):**

Up to three contextual cards like:

- "Plan my next working day" (Gmail + Calendar)
- "Find LinkedIn leads at a target company" (LinkedIn / Apify)
- "Draft a cold outreach message" (Gmail / Outlook / Instantly)
- "Analyze a Google Sheet"
- "Update my CRM after a call" (HubSpot / Salesforce / Pipedrive / Attio)
- etc.

Each card shows the integration logos it relies on. Clicking one fills the chat input with a starter prompt the user can edit before sending, and focuses the input so they can keep typing right away.

The cards never replace skills. If the agent already has skills, the existing skill-card empty state still wins — this only fills the gap where there was previously nothing.

## Why it matters

New users hit a blank chat and bounce. Existing users with newly connected integrations don't know what their agent can do with them. This turns both moments into a starting point.

The PR is also a step toward the broader "agents teach themselves to the user" direction — every new integration the team adds can add its own suggestion entries and start showing up here automatically.

## How it was verified

- Repro on the macOS desktop build: agent with no skills and no integrations renders the connect CTA; clicking it opens Integrations. Connecting Gmail and reopening the mission renders the Gmail-related suggestions; clicking a suggestion fills the input with the localized prompt and focuses the textarea.
- New unit suite for the suggestion derivation: 9 tests covering empty input, single-toolkit match, any-of toolkit logic, max cap, order preservation, custom max, slug normalization, and id uniqueness.
- `pnpm tsc --noEmit` green.
- `pnpm check-locales` clean across en / es / pt (15 namespaces in sync).

Repro video below.

### DIFF
--- a/app/src/components/toolkit-suggestions-empty-state.tsx
+++ b/app/src/components/toolkit-suggestions-empty-state.tsx
@@ -1,0 +1,59 @@
+import { useTranslation } from "react-i18next";
+import type { ToolkitSuggestion } from "../lib/toolkit-suggestions";
+import { SkillCard } from "./skill-card";
+
+interface Props {
+  suggestions: ToolkitSuggestion[];
+  onPick: (prompt: string) => void;
+  onConnectClick?: () => void;
+}
+
+export function ToolkitSuggestionsEmptyState({
+  suggestions,
+  onPick,
+  onConnectClick,
+}: Props) {
+  const { t } = useTranslation("board");
+  const isEmpty = suggestions.length === 0;
+  return (
+    <div className="self-stretch w-full h-full overflow-y-auto">
+      <div className="max-w-3xl mx-auto w-full px-6 pt-6 pb-4 flex flex-col gap-3">
+        <div className="text-center mb-1">
+          <h3 className="text-base font-semibold text-foreground">
+            {t(
+              isEmpty
+                ? "chatEmpty.suggestions.headingEmpty"
+                : "chatEmpty.suggestions.heading",
+            )}
+          </h3>
+          <p className="text-sm text-muted-foreground mt-1">
+            {t(
+              isEmpty
+                ? "chatEmpty.suggestions.subheadingEmpty"
+                : "chatEmpty.suggestions.subheading",
+            )}
+          </p>
+        </div>
+        {isEmpty && onConnectClick ? (
+          <SkillCard
+            image="electric-plug"
+            title={t("chatEmpty.suggestions.connectCta.title")}
+            description={t("chatEmpty.suggestions.connectCta.description")}
+            onClick={onConnectClick}
+          />
+        ) : (
+          suggestions.map((s) => (
+            <SkillCard
+              key={s.id}
+              image={s.image}
+              title={t(`chatEmpty.suggestions.${s.id}.title` as const)}
+              description={t(`chatEmpty.suggestions.${s.id}.description` as const)}
+              integrations={s.toolkits}
+              onClick={() => onPick(t(`chatEmpty.suggestions.${s.id}.prompt` as const))}
+            />
+          ))
+        )}
+      </div>
+    </div>
+  );
+}
--- a/app/src/components/use-agent-chat-panel.tsx
+++ b/app/src/components/use-agent-chat-panel.tsx
@@ -35,7 +35,9 @@ import {
   type UserAttachmentMessageLabels,
 } from "@houston-ai/chat";
 
+import { useDraftStore } from "../stores/drafts";
 import { useFeedStore } from "../stores/feeds";
+import { useUIStore } from "../stores/ui";
 import { useWorkspaceStore } from "../stores/workspaces";
 import {
   useConnectedToolkits,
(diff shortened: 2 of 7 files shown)


emutis21:
https://github.com/user-attachments/assets/560151cc-147f-4f78-bb2b-df6116131c48

ja-818:
Thanks Esteban — this one is genuinely good work. The user research is real (the WhatsApp thread on May 9), the i18n is comprehensive across en / es / pt, the derivation function has 9 well-chosen tests, the pure-data vs rendering split is clean, and you reused `SkillCard` + `useConnectedToolkits` instead of reinventing. The shape is right.

A handful of things to tighten before merge though — none of them are architectural, mostly polish.

## Issues worth fixing before merge

### 1. The empty-state branch is wrong for \"connected but no match\"
```ts
const isEmpty = suggestions.length === 0;
// → shows \"Connect a tool to get started\"
```
If a user has connected something we don't have a suggestion for (Twilio, some niche toolkit, etc.), `suggestions.length === 0` but the user already connected a tool. The UI then tells them to \"Connect a tool to get started\" — wrong message.

Split the condition: `connectedSet.size === 0` → connect CTA, else if `suggestions.length === 0` → a generic \"just type what you need\" fallback.

### 2. `as const` in `t()` calls bypasses i18n type checking
```tsx
t(`chatEmpty.suggestions.${s.id}.title` as const)
```
Our CLAUDE.md i18n rule is explicit: keys are type-checked via the `react-i18next.d.ts` augmentation so typos fail at compile time. This pattern defeats that. If someone adds an entry to `TOOLKIT_SUGGESTIONS` and forgets to add the matching locale keys, TS won't catch it — the user just sees the literal key string rendered as text.

Either inline a small switch on `s.id` (8 cases, no big deal), or assert at compile time that each id maps to its three keys.

### 3. DOM querySelector to focus the textarea is fragile
```tsx
requestAnimationFrame(() => {
  document.querySelector<HTMLTextAreaElement>('textarea[name=\"message\"]')?.focus();
});
```
Two issues:
- **Cross-package coupling via attribute selector.** The `name=\"message\"` lives inside `@houston-ai/chat`'s `prompt-input.tsx` — it's an internal markup detail of a library you don't control, not a public contract. If chat ever renames it, this silently breaks with no error surface.
- **First-match selector.** `document.querySelector` returns the first textarea in DOM order, not the active panel. Houston can have multiple chat surfaces mounted at once (Try tutorial, dashboard, board). It works on board-tab today but it's incidental.

The clean fix is a forwarded ref through `AIBoard` / `ChatPanel`, or a `useImperativeHandle` exposing `.focus()`. Happy to pair on the shape.

### 4. `requestAnimationFrame` is a one-shot
No retry, no fallback. If the textarea isn't in the DOM after one frame (slow render, strict-mode double-render), focus silently fails. Combined with the imperative draft-store write, the whole click → input-filled → cursor-in-textarea chain has 3 silent-failure points and per our no-silent-failures rule, at least one of them should warn or surface.

### 5. Hardcoded toolkit slugs with no coupling to source of truth
`TOOLKIT_SUGGESTIONS` hardcodes `\"linkedin\"`, `\"apify\"`, `\"gmail\"`, etc. If Composio renames a slug upstream, suggestions stop firing silently — no error, no log. The lowercase test is a self-check against a hardcoded list, not against the toolkit registry.

Add a test that intersects these slugs against the actual toolkit list — that turns a silent regression into a CI failure.

## Quality nits

### 6. Hardcoded priority is sales-biased
LinkedIn leads is #1 in the array. A user who has connected Gmail + Notion + Sheets only gets sales-flavored suggestions if LinkedIn-OR-Apify happens to be in there too, because of fixed array order. Not a blocker, but worth thinking about ranking by number of matched toolkits, or by recency of connection, instead of array index.

### 7. No tests for the component or the side-effect chain
The derivation function is well-tested, but `ToolkitSuggestionsEmptyState` rendering and the draft-store + focus side-effect — the parts most likely to break — have zero coverage. At minimum: a test that clicking a card calls `setDraftText` with the right session key.

### 8. `\"new-conversation\"` magic string
Already used across `ai-board.tsx`, `drafts.ts`, `board-tab.tsx`, so you're not introducing it. But while we're touching this area, worth extracting to a shared constant in `app/src/stores/drafts.ts`. Small follow-up if you want.

## Suggested rework

Half a day at most:
1. Fix the empty-state branch to check `connectedSet.size` (not suggestions length).
2. Replace `as const` t() with a typed lookup or inline switch.
3. Replace the `document.querySelector` focus with a forwarded ref / imperative handle.
4. Add a component test for the click → draft → focus chain.

The other items can land as a follow-up, but those four would get this to a state I'd merge confidently.

Happy to pair on the focus-ref refactor if it'd help — that's the only one with a bit of plumbing.

PR #113, opened 2026-05-11
feat(chat): paste images from OS clipboard

cravenceiling:
## Summary

Adds image paste from the OS clipboard into the chat/board composer, and
hardens the implementation following maintainer review.

Pasting an image (screenshot, "Copy Image", Figma/Slack, etc.) into the
composer now attaches it as a file, alongside the existing drag-drop and
file-picker paths. The same dedupe/validation pipeline is reused so all
three entry points behave identically.

## Scope decision: Wayland

The native clipboard bridge from the first iteration (arboard + a Tauri
command pulling ~10 transitive crates) was removed entirely. The path is
the web `ClipboardEvent`, which covers macOS WebView, Windows WebView2,
and Linux **X11** WebKitGTK.

Linux **Wayland** WebKitGTK does not surface image clipboards to the web
layer. Native Wayland support is **deliberately deferred** off this
branch. It is **not silent**: the detectable case (webkitgtk lists an
image item but `getAsFile()` returns null) raises a visible toast telling
the user to drag the file in instead. The undetectable image-only case
(nothing reaches JS) is documented at the source as a known follow-up.

## Changes

- `resolveClipboardPaste` returns `files | unavailable | ignore`; text
  paste falls through untouched.
- Clipboard dedupe key now includes `file.type`, so two distinct unnamed
  images sharing size + mtime no longer collapse into one. The shared
  file-identity key (drop/picker) is unchanged.
- `ensureFileName` strips MIME parameters/suffixes from the generated
  extension; unclean subtypes fall back to `bin`.
- Extracted `useAttachmentIntake` (shared ingest core, now used by both
  `ChatInput` and `ChatPanel`, removing duplicated logic) and
  `useComposerAttachments` (composer picker/paste surface). `chat-input`
  is back under the 200-line limit.
- `ChatComposerLabels` moved to `chat-panel-types` (types module no
  longer depends on a component); added `imagePasteUnavailable`.
- i18n: fixed singular/plural drop-overlay copy across en/es/pt and the
  `ui/` defaults; new label wired through `chat-tab` and `board-tab`.

## Test plan

- [x] `node --test ui/chat/test/clipboard-files.test.mjs` — 19/19 pass
  (null/undefined input, type-disambiguated dedupe both directions,
  parametrized-MIME extension, `clipboardSignalsImage`, every
  `resolveClipboardPaste` outcome).
- [x] `pnpm typecheck` (ui workspace) — clean.
- [x] `pnpm check-locales` (app) — locales in sync, no em-dashes.
- [x] Manual: paste a screenshot on macOS / Windows / Linux-X11 — image
  attaches.
- [ ] Manual: paste on Linux-Wayland — visible toast, no silent no-op;
  drag-drop still works.
- [x] Manual: paste plain text — still goes into the textarea.


ja-818:
Thanks for the PR Daniel. The feature works end-to-end and the tests on the TS side are a nice touch, but I want to push back on the implementation before we merge — there are a few things here that don't meet the bar we hold internal code to. None of them are nitpicks; most have concrete project-rule citations.

## Blockers

### 1. Sync Tauri command + no timeout on the Linux subprocess
`pub fn read_clipboard_image()` is synchronous, and the Linux fallback uses `std::process::Command::output()` with no timeout and no `kill_on_drop`. Our internal protocol is explicit: any subprocess invoked from a Tauri command must use `tokio::process::Command` with a timeout and `kill_on_drop(true)` (we got burned by a hang here before). A stuck Wayland compositor or a hung `wl-paste` will deadlock the worker indefinitely. Re-encoding a 4K screenshot's RGBA buffer to PNG also blocks the IPC thread — that needs `spawn_blocking` or an `async fn`.

### 2. Wrong scope / wrong primary path
The web `ClipboardEvent` already exposes pasted image data on macOS WebView, Windows WebView2, and X11 WebKitGTK. Your `e.clipboardData` path in `chat-input.tsx` handles those cases for free. The only platform that genuinely needs a native bridge is **Linux Wayland WebKitGTK**, because webkitgtk doesn't surface image-only clipboards under Wayland.

But the PR pulls `arboard` as a cross-platform fallback for every OS, and `arboard` drags in ~10 transitive crates (`image`, `tiff`, `half`, `moxcms`, `pxfm`, `x11rb`, `gethostname`, `weezl`, `zune-jpeg`, `png 0.18`). That's a non-trivial compile-time + binary-size + attack-surface cost for a fallback that's effectively dead code on macOS and Windows. The shape we'd want is: web event handles 99% of cases, native bridge is Linux-only behind `#[cfg(target_os = \"linux\")]`.

### 3. Silent error path
```ts
readClipboardFiles().then(...).catch((err) => { onNotice?.(err…); });
```
`onNotice` is optional, so if a caller doesn't pass it the error vanishes. Meanwhile `app/src/lib/clipboard.ts` already calls `showErrorToast` before returning `[]`. One path toasts, the other swallows. We have an explicit no-silent-failures policy — every error a user-initiated action can produce has to reach the user as a visible toast with a Report-bug affordance. Pick one layer to surface from and drop the other.

## Quality issues

### 4. Pointless decode → re-encode
`arboard::get_image()` decodes whatever the clipboard contains into raw RGBA, and then `encode_png` re-encodes it to PNG. The original PNG bytes (the common case for screenshots) get thrown away. Wasteful on large captures.

### 5. Base64-over-string IPC for image bytes
Tauri 2 supports binary IPC via `tauri::ipc::Response`. A 5 MB screenshot becomes a 6.6 MB UTF-8 string, then serde-json escapes it, then the frontend does `atob` plus a byte-by-byte `charCodeAt` copy. Either write to a tempfile and pass the path, or use a typed binary channel.

### 6. No size limit
Nothing in `read_clipboard_image` rejects multi-hundred-MB clipboards. User-initiated, but trivially abusable by a misbehaving source app and an easy way to OOM the renderer.

### 7. Two `png` crate versions in Cargo.lock
You add direct `png = \"0.17\"`, but `arboard → image` pulls `png = \"0.18\"`. Both get compiled. Use the one already in the tree, or pull the encoder out of the `image` crate.

### 8. Logic bug in `shouldReadNativeClipboardFiles` for mixed clipboards
Pasting from Figma or Slack drops a clipboard with both a `\"string\"` item and an image. The current code short-circuits on \"any item is kind === 'string'\" before checking for the file, so on Linux Wayland (where the web layer doesn't expose the image) the paste becomes a silent no-op. Check for file/image presence first, then bail on text-only.

### 9. JPEG silently unsupported on Wayland
The Linux fallback only asks for `image/png`. Firefox \"Copy Image\" of a JPEG ends up as empty stdout → `Ok(None)` → falls through to arboard, which may or may not work depending on the session. Either iterate MIME types, or query `wl-paste --list-types` and pick.

### 10. `.unwrap_or(0)` on `SystemTime`
Our internal banned-patterns list explicitly calls out `.unwrap_or(...)` over a user-initiated op. Impact is minor here (just a filename), but it's the kind of corner that adds up.

### 11. No Rust tests
The TS test for `clipboard-files.ts` is good, but the Rust side has zero coverage — no test for the empty-clipboard path returning `Ok(None)`, no test for the Linux fallback ordering, no test for the size cap (once it's added).

### 12. `png 0.17` hardcodes `BitDepth::Eight`
Edge case (HDR / Display P3 screenshots on newer macOS) but it silently truncates 16-bit-per-channel images.

## Suggested rework

Scope this down to what it actually is — a Linux Wayland workaround:

1. Delete the `arboard`, `png`, `base64` deps. Use the web `ClipboardEvent` as the only path on macOS, Windows, and Linux-X11. Verify with manual paste on each.
2. Implement the native bridge behind `#[cfg(target_os = \"linux\")]`. Make the Tauri command `async fn`. Use `tokio::process::Command` with a ~2s timeout and `kill_on_drop(true)`.
3. Query types first (`wl-paste --list-types`), pick the best image MIME, return raw bytes via `tauri::ipc::Response`. Add a size cap (50 MB feels right).
4. Drop the `onNotice` swallow in `chat-input.tsx` — toast at one layer, not two.
5. Fix the mixed-clipboard short-circuit in `shouldReadNativeClipboardFiles`.
6. Add Rust tests: empty clipboard, oversize, content-type fallback ordering.

That's roughly half a day of rework and it turns this into a feature we can confidently ship to non-technical users on every platform we support.

Happy to pair on any of these if it'd help.

PR #233, opened 2026-05-21
feat(routines): cancel in-flight runs + cascade on delete

### LINKED ISSUE #232: Routines: no way to stop an in-flight run; delete-routine doesn't cascade
## Reported behaviour

User runs a routine manually (Anthropic provider). Mid-run, their Claude token quota exhausts. The `claude` CLI subprocess blocks internally on the rate-limit window and resumes silently after reset, burning more tokens than the routine normally uses.

User attempts to stop it:

- No stop / cancel affordance in the routines UI for an in-flight run.
- Deleting the routine has no effect on the run already in flight.
- Logging out of Anthropic has no effect (credentials are already loaded into the subprocess).
- Closing the app window leaves the engine subprocess running.
- Only Force Quit (process tree kill) stops the work.

## Root cause

1. The routines layer ships **no cancellation primitive**. `run_routine` (`engine/houston-engine-core/src/routines/runner.rs`) is fire-and-await. The session it spawns has a stable `session_key` and the engine already has `sessions::cancel` (used by chat), but the routines API never exposes it.
2. `routines::delete` mutates JSON and aborts the cron job via `scheduler.sync_agent`. It does **not** cancel any in-flight `run_routine` task spawned for that routine — the detached `tokio::spawn` and its child CLI subprocess keep going.
3. The `RoutinesTab` UI only wires `useRunRoutineNow`. There is no `useCancelRoutineRun`, no stop button on a `running` row.

## Out of scope for this issue

- The CLI's silent sleep-through-rate-limit behaviour (Anthropic CLI internal). With a cancel button the user can interrupt; we don't need to teach Houston about the rate-limit window for this fix.
- App close not propagating to the engine subprocess (engine supervisor concern, separate).
- Logout not killing in-flight runs (creds live in the subprocess; would need a different approach).

## Acceptance

- A user can stop an in-flight routine run from the routines UI.
- Deleting a routine cancels any in-flight runs of that routine before the routine itself is removed.
- Cancelled runs end in a distinct terminal state (`cancelled`) so history shows what happened, not a misleading `error`.
- Race-safe: a cancel that lands while the dispatcher is mid-flight wins; the runner doesn't overwrite `cancelled` with `error`/`silent`/`surfaced` and doesn't create a board activity.

cravenceiling:
## Summary

- Add cancellation end-to-end for in-flight routine runs (new `cancel` action route + UI Stop button).
- `DELETE /v1/routines/:id` now cascades to any `running` runs of that routine, terminating the provider subprocess and marking the run `cancelled`.
- New terminal status `cancelled` (distinct from `error`) with race-safe write ordering — a cancel that lands mid-dispatch wins; the runner never overwrites it with `silent`/`surfaced`/`error` and never surfaces a board activity for a cancelled run.

Closes #232.

## What changed

**Engine**
- `engine/houston-engine-core/src/routines/runs.rs` — `find_by_id` helper.
- `engine/houston-engine-core/src/routines/runner.rs` — `cancel_run(rt, events, root, agent_path, run_id)` core primitive + race-protection branch in `run_routine`.
- `engine/houston-engine-server/src/routes/routines.rs` — `POST /v1/routines/:id/runs/:run_id:cancel` route. `DELETE /v1/routines/:id` cancels in-flight runs before deleting.
- `engine/houston-engine-core/src/routines/types.rs` — doc the new `cancelled` status.
- `ui/agent-schemas/src/routine_runs.schema.json` — `cancelled` in the enum.

**Frontend**
- `ui/engine-client/src/client.ts` — `cancelRoutineRun(agentPath, routineId, runId)`.
- `ui/routines/src/run-history.tsx` — per-row Stop button when status is `running`, new `Cancelled` status pill, accepts `onCancelRun` prop.
- `ui/routines/src/routine-editor.tsx` — header `Run now` button swaps to `Stop` while a run is in flight, threads `onCancelRun` into `RunHistory`.
- `ui/routines/src/types.ts` — `RunStatus` adds `cancelled`.
- `app/src/lib/tauri.ts` + `app/src/hooks/queries/use-routines.ts` + `app/src/components/tabs/routines-tab.tsx` — `tauriRoutines.cancelRun`, `useCancelRoutineRun` mutation, wiring.

**Tests**
- 6 new unit tests in `houston-engine-core::routines::*` covering `find_by_id`, `cancel_run` happy/conflict/missing, race-protection branch.
- 2 new integration tests in `engine/houston-engine-server/tests/routines.rs` covering the cancel route (200/409/400/404) and the delete-cascade.

**Docs**
- `knowledge-base/engine-protocol.md` — new route documented.

## Test plan

- [x] `cargo test -p houston-engine-core --lib routines::` → 26 passing (6 new).
- [x] `cargo test -p houston-engine-server --test routines` → 5 passing (2 new).
- [x] `cd app && pnpm typecheck` → clean.
- [x] `cd app && pnpm check-locales` → clean.
- [x] Start an Anthropic-backed routine via Run now. While it's running, click Stop. Verify the run row flips to `Cancelled — Stopped by user`, the provider subprocess is gone, no board activity created.
- [x] Start a routine via Run now. While running, delete the routine. Verify the in-flight run flips to `cancelled` and the subprocess is gone.
- [ ] Try to cancel an already-terminal run via the API → 409.

## Out of scope (intentional)

- The CLI's silent **pause-and-resume on Anthropic usage limit** — the underlying reason the original user's routine ran far longer than usual. Cancel gives them the off-switch; surfacing the paused state ("Paused — resumes at HH:MM") is a separate piece of work. **A follow-up commit on this branch will start that surface so the user sees *why* a routine is taking long.**
- App-window-close not propagating to the engine subprocess (engine supervisor concern, separate).
- Logout-while-running not invalidating in-flight runs (creds live in the subprocess at spawn time).

### DIFF
--- a/app/src/components/tabs/routines-tab.tsx
+++ b/app/src/components/tabs/routines-tab.tsx
@@ -9,6 +9,7 @@ import {
   useUpdateRoutine,
   useDeleteRoutine,
   useRunRoutineNow,
+  useCancelRoutineRun,
 } from "../../hooks/queries";
 import { useTimezonePreference } from "../../hooks/use-timezone-preference";
 import type { TabProps } from "../../lib/types";
@@ -57,6 +58,7 @@ export default function RoutinesTab({ agent }: TabProps) {
   const updateRoutine = useUpdateRoutine(path);
   const deleteRoutine = useDeleteRoutine(path);
   const runNow = useRunRoutineNow(path);
+  const cancelRun = useCancelRoutineRun(path);
 
   const [view, setView] = useState<View>({ type: "grid" });
   const [form, setForm] = useState<RoutineFormData>(EMPTY_FORM);
@@ -146,6 +148,13 @@ export default function RoutinesTab({ agent }: TabProps) {
     [runNow],
   );
 
+  const handleCancelRun = useCallback(
+    (routineId: string, runId: string) => {
+      cancelRun.mutate({ routineId, runId });
+    },
+    [cancelRun],
+  );
+
   // `useTimezonePreference` auto-seeds on first call, so `tz.timezone` is
   // non-null from the first render. We still wait for the roundtrip to
   // finish so the cron schedule renders against the real zone instead of
@@ -175,6 +184,11 @@ export default function RoutinesTab({ agent }: TabProps) {
         routine={editing}
         runs={editingRuns}
         onRunNow={editing ? () => handleRunNow(editing.id) : undefined}
+        onCancelRun={
+          editing
+            ? (runId: string) => handleCancelRun(editing.id, runId)
+            : undefined
+        }
         onToggle={
           editing ? (enabled) => handleToggle(editing.id, enabled) : undefined
         }
--- a/app/src/components/tabs/routines-tab.tsx
+++ b/app/src/components/tabs/routines-tab.tsx
@@ -184,6 +184,7 @@ export default function RoutinesTab({ agent }: TabProps) {
         routine={editing}
         runs={editingRuns}
         onRunNow={editing ? () => handleRunNow(editing.id) : undefined}
+        runNowPending={runNow.isPending}
         onCancelRun={
           editing
             ? (runId: string) => handleCancelRun(editing.id, runId)
--- a/app/src/hooks/queries/index.ts
+++ b/app/src/hooks/queries/index.ts
@@ -20,6 +20,7 @@ export {
   useUpdateRoutine,
   useDeleteRoutine,
   useRunRoutineNow,
+  useCancelRoutineRun,
 } from "./use-routines";
 export {
   useLearnings,
--- a/app/src/hooks/queries/use-routines.ts
+++ b/app/src/hooks/queries/use-routines.ts
@@ -69,3 +69,14 @@ export function useRunRoutineNow(agentPath: string) {
     },
   });
 }
+
+export function useCancelRoutineRun(agentPath: string) {
+  const qc = useQueryClient();
+  return useMutation({
+    mutationFn: ({ routineId, runId }: { routineId: string; runId: string }) =>
+      tauriRoutines.cancelRun(agentPath, routineId, runId),
+    onSuccess: () => {
+      qc.invalidateQueries({ queryKey: ["routine-runs", agentPath] });
+    },
+  });
+}
--- a/app/src/lib/tauri.ts
+++ b/app/src/lib/tauri.ts
@@ -547,6 +547,10 @@ export const tauriRoutines = {
     call<void>("run_routine_now", () =>
       getEngine().runRoutineNow(agentPath, routineId),
     ),
+  cancelRun: (agentPath: string, routineId: string, runId: string) =>
+    call("cancel_routine_run", () =>
+      getEngine().cancelRoutineRun(agentPath, routineId, runId),
+    ),
   startScheduler: (agentPath: string) =>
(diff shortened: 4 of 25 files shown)


cravenceiling:
Follow-up commit pushed: 5f9d293 — surface usage-limit pause so users know a run is waiting.

Addresses the visibility side of the originally reported scenario. The reporter saw a routine "running" for hours with no explanation; root cause is that claude-code sleeps internally on its plan-window usage limit instead of exiting, so Houston's session loop never saw a terminal signal.

What changed:
- New non-terminal ProviderError::UsageLimitPaused variant with resets_at hint.
- Anthropic classifier matches the "Claude usage limit reached. Your limit will reset at HH:MM (TZ)" banner BEFORE the existing QuotaExhausted branch.
- session_runner gains a generic SessionLifecycle hook (no routines dependency) that fires on_paused/on_resumed.
- Routine dispatcher persists routine_run.paused_until and emits RoutineRunsChanged. Cleared when output resumes.
- UI: amber dot + "Waiting · resumes at {hint}" on RoutineRow; "Waiting for usage limit — resumes at {hint}" in RunHistory.
- Stop button still works on a paused run — SIGTERM kills a sleeping subprocess.

Out of scope intentionally: token-cost estimation during pause, auto-cancel after N hours, paused card in the chat panel.

cravenceiling:
Third commit pushed: a0886f2 — fix(routines): guard run-now against spam-clicks, sweep orphans, route delete-cascade.

Closes #236 in addition to #232.

Triage of a user report turned up three bugs that were either pre-existing or shipped half-fixed in this PR:

1. **Run Now had no spam-click guard.** Clicking 4 times in 28s during a 1m23s codex run produced 1 real "surfaced" row + 3 garbage "error: conflict: another mission..." rows because `try_acquire_workdir` fired inside the dispatcher AFTER the routine_run was already created. New precondition in `run_routine` rejects with 409 before touching disk. Frontend disables the Run Now button (`Starting…`) while the mutation is pending to close the propagation race.

2. **Orphan "running" rows from a crashed engine would block all future run-now once (1) was in place.** Added `runs::sweep_orphan_running` — runs once per agent at scheduler-start, reaps stale `status=running` rows to `status=error` with summary "Engine restarted before this run finished". Idempotent.

3. **Delete-cascade was on the wrong route.** This PR added the cancel-on-delete cascade to `/v1/routines/:id` (`routes/routines.rs::remove`), but the frontend `engine-client::deleteRoutine` actually calls `/v1/agents/routines/:id` (`routes/agents.rs::delete_routine`). Mirrored the cascade there too — until we consolidate the two routes, both paths need it.

Tests:
- New unit test: `run_routine_rejects_when_another_run_is_in_flight`.
- New unit tests: `sweep_orphan_running_marks_only_running_rows`, `sweep_orphan_running_is_noop_when_clean`.
- New integration test: `run_now_returns_409_when_another_run_is_in_flight` (verifies the route 409s AND that no new row is created).

`cargo test -p houston-engine-core --lib routines::` → 33 passing.
`cargo test -p houston-engine-server --test routines --test agents` → 10 passing.
TS typechecks + locales clean.

Out of scope for this PR (good cleanup for a separate one): the duplicate `routines` and `routine_runs` modules in engine-core (`crate::routines` vs `crate::agents::routines`) should be unified — they touch the same JSON files but expose subtly different APIs, and `agents::types::RoutineRun` is missing the `paused_until` field added in the previous commit, which means the GET path silently drops the hint on the round trip.

PR #242, opened 2026-05-22
chore(ci): add PR-time workflow (typecheck + cargo check + tests)

broomva:
Tracks #243 — RFC: bstack primitives + control-metalayer.

---

## Summary

Adds `.github/workflows/ci.yml` that automates the contributor checks already documented in CONTRIBUTING.md plus the i18n parity check from `knowledge-base/i18n.md`. Closes the no-PR-CI gap (`release.yml` and `engine-release.yml` are tag-triggered only — nothing runs on PRs today).

Two parallel jobs on `ubuntu-latest`, plus a `CI Success` aggregator so branch protection can point at a single required check:

- **TypeScript**: `pnpm install` → `pnpm -r typecheck` → `cd app && pnpm check-locales`
- **Rust**: apt-get Tauri 2 deps → `cargo build --release -p houston-engine-server` → `cargo check --workspace --all-targets` → `cargo test --workspace`
- **CI Success**: depends on both, fails red when either fails real (not when cancelled)

## Why the engine-sidecar build step

`app/src-tauri/build.rs:29` invokes `tauri_build::build()` unconditionally on every cargo invocation. `tauri-build` reads `externalBin: ["binaries/houston-engine"]` from `tauri.conf.json:47` and copies the resolved binary into the bundle. The stager at `build.rs::stage_engine_sidecar` falls back to `target/release/houston-engine` when no `--target` is set (build.rs:158) — so a plain `cargo build --release -p houston-engine-server` populates that path and satisfies tauri-build's externalBin requirement for both `cargo check` and `cargo test` downstream.

Without this step, `cargo check --workspace` fails on first CI run because `tauri-build` returns `Error::ResourcePathNotFound` for the missing externalBin (`build.rs::stage_engine_sidecar`'s `cargo:warning` fallback at build.rs:14-16 doesn't skip `tauri_build::build()` — only the staging is non-fatal).

## Convention compliance

- Conventional commit `chore(ci):` per CONTRIBUTING.md §"Commit Messages"
- Action versions mirror `release.yml` + `engine-release.yml`: `actions/checkout@v4`, `pnpm/action-setup@v4`, `actions/setup-node@v4` with `node-version: 20` + `cache: pnpm`, `dtolnay/rust-toolchain@stable`, `actions/cache@v4`
- Cache key `${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}` with restore-keys `${{ runner.os }}-cargo-` — shares the `<os>-cargo-` prefix with `release.yml:190` (`<os>-cargo-universal-`) and `release.yml:804` (`<os>-cargo-<arch>-`), so PR CI partial-restores from warm release caches and vice versa
- `permissions: contents: read` — CI never writes
- Concurrency cancels stale PR runs but never main pushes (so every main commit gets a definitive green/red for release/rollback decisions)
- Aggregator uses `if: ${{ !cancelled() }}` (not `if: always()`) — cancelled PRs short-circuit the aggregator instead of leaving phantom red checks on older SHAs

## Out of scope (each is a self-contained follow-up)

- **`cargo fmt --check` + `cargo clippy`**. Local `cargo fmt --all -- --check` against current main reports extensive drift across many crates (verifiable locally). Landing a `cargo fmt --all` baseline is itself a substantial diff and belongs in its own PR. CLAUDE.md §"No silent failures" banned patterns (`let _ = <fallible>`, `.ok()` to drop a Result, log-and-continue match arms) are the obvious first home for the initial clippy lint set once fmt baseline lands.

- **Typecheck script for `@houston-ai/agent-schemas`**. That package has no `scripts` block (`ui/agent-schemas/package.json:1-17`), so `pnpm -r typecheck` silently skips it. Adding the script touches a different package surface area and was kept out to keep this PR strictly focused on the CI-gap fix.

- **Full Tauri build (macOS DMG + Windows MSI)** stays in `release.yml` on tag push. PR CI does not need release artifacts.

## Test plan

- [ ] GitHub YAML linter on push validates the workflow file
- [ ] CI on this very PR is itself the smoke test — if it goes green, the workflow is sound
- [ ] After merge, the next PR will run the same checks against its diff

## Notes for maintainer

- Branch protection can wire the single `CI Success` check as the required gate
- If the apt deps list needs additions (Tauri 2 evolves), updates can be a one-line edit to the workflow
- Sidecar build adds ~5-7 min cold to CI; cache key cross-pollination with release.yml reduces warm-cache cost

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,271 @@
+# Houston PR CI Workflow
+#
+# Runs on every PR (any target branch — supports stacked PR workflows
+# off claude/<worktree-name> branches per CLAUDE.md §Git) and every
+# push to main. Enforces the contributor checks already listed in
+# CONTRIBUTING.md (`pnpm typecheck`, `cargo check --workspace`,
+# `cargo test --workspace`) plus the i18n parity check from
+# knowledge-base/i18n.md (`pnpm check-locales` — en/es/pt key +
+# placeholder + em-dash drift).
+#
+# Two parallel jobs run on ubuntu-latest; a final `CI Success`
+# aggregator gives maintainers a single required check to point branch
+# protection at. The Tauri app release build (macOS .app + Windows
+# .msi) stays in release.yml — PR CI doesn't produce release artifacts,
+# but it DOES run app/src-tauri's build.rs as part of
+# `cargo check --workspace`, which is why the Rust job stages the
+# engine sidecar (`cargo build --release -p houston-engine-server`)
+# before checks: `tauri-build` reads the `externalBin` entry in
+# tauri.conf.json:47 (`binaries/houston-engine`) unconditionally on
+# every cargo invocation, and the stager at app/src-tauri/build.rs
+# falls back to `target/release/houston-engine` when no --target is
+# set.
+#
+# Action-version conventions match release.yml and engine-release.yml:
+#   actions/checkout@v4, pnpm/action-setup@v4 (auto-reads packageManager
+#   pin from root package.json), actions/setup-node@v4 with node 20 +
+#   cache: pnpm, dtolnay/rust-toolchain@stable, actions/cache@v4.
+#
+# `cargo fmt --check` and `cargo clippy` are intentionally NOT gating
+# in this initial workflow because a local `cargo fmt --check` against
+# current main reports extensive drift across many crates (engine
+# subcrates + app/src-tauri); verifiable by running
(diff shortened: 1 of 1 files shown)


broomva:
Consolidated into #272 — one reviewable 'agent-driven dev workflow' proposal (CI + governance + dogfood docs + banned-patterns), tied to RFC #243/#251. Review/accept/reject as a unit there. Closing this one.

Issue #303, opened 2026-05-26
[BUG] Duplicated response


cravenceiling:
After getting the first response and sending a new message I changed to other agent and then returned to the previous chat and now I see the same response and below it the new one.

<img width="610" height="598" alt="Image" src="https://github.com/user-attachments/assets/a8c3a8d5-4d51-4ef2-8f6b-eac13f65f7cf" />

PR #370, opened 2026-06-01
fix(chat): dedupe hydrated feed so surfaced routines don't double their first turn (#363)

### LINKED ISSUE #363: Bug: Routines when they surface to activity show always the first message from the user and the AI repeated, this needs to be fixed

cravenceiling:
## Issue

Fixes #363 — *Routines, when they surface to activity, always show the first message from the user and the AI repeated.*

## Root cause

A routine runs in the background; its transcript streams to the desktop feed store over WS (under `routine-{rid}-run-{run}`) and is persisted to the DB. When the surfaced activity is opened, `handleHistoryLoaded` merges the persisted **DB history** with the **live in-memory feed bucket**.

That merge used exact `JSON.stringify` equality to drop duplicates:

```ts
const serverIds = new Set(items.map((it) => JSON.stringify(it)));
const tail = current.filter((it) => !serverIds.has(JSON.stringify(it)));
setFeed(path, sessionKey, [...items, ...tail]);
```

Exact equality cannot recognize a live `assistant_text_streaming` item as the **same turn** as the persisted `assistant_text` final (different `feed_type`), so both survive the merge and the first user message + AI response render **twice**.

This is a regression: the fix already existed as `mergeFeedHistory` on the `claude/fix-issue-303-duplicated-response` branch (commit `2c29e8a`, "fix(chat): dedupe hydrated feed items") but was never merged to `main` and was lost. #363 is the same bug (#303) resurfacing on routines.

## Fix (restored + tested)

- **`ui/chat/src/feed-merge.ts`** — restore `mergeFeedHistory(history, current)`: dedupes the live bucket against history by **exact key** *and* **streaming/final equivalence** (`assistant_text`↔`assistant_text_streaming`, `thinking`↔`thinking_streaming`), using **count-based** consumption so a genuinely repeated turn still survives. `mergeFeedItem` is hardened to drop an exact / stream-final-equivalent duplicate within the current turn (`replaceLast` is now nullable).
- **`ui/chat/src/index.ts`** — export `mergeFeedHistory`.
- **`app/src/components/tabs/board-tab.tsx`** + **`app/src/components/use-mission-control.ts`** — both `handleHistoryLoaded` sites now call `mergeFeedHistory(...)` instead of the naive `JSON.stringify` tail merge.
- **`ui/chat/test/feed-merge.test.mjs`** — regression tests: surfaced-routine stale-stream case, equal-history case, live-tail preservation, repeated-user-turn preservation, plus the hardened `mergeFeedItem` cases.

Scope is kept to the duplication (#363). The sibling "Mission Control can't load routine history" gap (forced `activity-{id}` in `conversations.rs`) is a separate issue (#364) and is **not** touched here.

## Affected files
- `ui/chat/src/feed-merge.ts`
- `ui/chat/src/index.ts`
- `ui/chat/test/feed-merge.test.mjs`
- `app/src/components/tabs/board-tab.tsx`
- `app/src/components/use-mission-control.ts`

## Test
```bash
# unit (ui/chat) — the regression coverage
node --test ui/chat/test/feed-merge.test.mjs

# typecheck
pnpm -r typecheck
cd app && pnpm tsc --noEmit
```

### DIFF
--- a/app/src/components/tabs/board-tab.tsx
+++ b/app/src/components/tabs/board-tab.tsx
@@ -3,7 +3,7 @@ import { useTranslation } from "react-i18next";
 import { useQueryClient } from "@tanstack/react-query";
 import { AIBoard } from "@houston-ai/board";
 import type { KanbanItem, NewPanelOpener } from "@houston-ai/board";
-import type { FeedItem } from "@houston-ai/chat";
+import { mergeFeedHistory, type FeedItem } from "@houston-ai/chat";
 import { Terminal, GitBranch } from "lucide-react";
 
 import { useFeedStore } from "../../stores/feeds";
@@ -239,17 +239,9 @@ export default function BoardTab({ agent, agentDef }: TabProps) {
       // truth — live WS events append cleanly and no "liveFeed wins if
       // non-empty" hack is needed. Any items already in the bucket
       // from WS events that arrived between activity creation and
-      // selection are preserved by merging the server history with
-      // the current bucket and dropping exact duplicates by position.
+      // selection are preserved by the shared history merge.
       const current = useFeedStore.getState().items[path]?.[sessionKey] ?? [];
-      // Server history is authoritative for everything persisted up to
-      // load time. Anything currently in `current` that isn't in the
-      // server history must be either an optimistic overlay we pushed
-      // or an event that landed mid-load. Append those after the
-      // server slice.
-      const serverIds = new Set(items.map((it) => JSON.stringify(it)));
-      const tail = current.filter((it) => !serverIds.has(JSON.stringify(it)));
-      setFeed(path, sessionKey, [...items, ...tail]);
+      setFeed(path, sessionKey, mergeFeedHistory(items, current));
     },
     [path, setFeed],
   );
--- a/app/src/components/use-mission-control.ts
+++ b/app/src/components/use-mission-control.ts
@@ -1,7 +1,7 @@
 import { useState, useCallback, useMemo, useRef } from "react";
 import { useTranslation } from "react-i18next";
 import type { KanbanItem } from "@houston-ai/board";
-import type { FeedItem } from "@houston-ai/chat";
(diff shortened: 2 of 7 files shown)


cravenceiling:
## Follow-up: the duplicate is the USER message, fixed in 7214f3c

Testing with the first commit (rebuilt) showed the **user message** duplicated, not the AI reply — `[user, "pong", user(dup), <empty Mission log>]`. The first commit (`mergeFeedHistory`) only deduped the *assistant* streaming/final pair, so it didn't cover this.

**Confirmed root cause (client-side):** the engine persists the routine's user prompt exactly once (`opts.user_message.take()`; a fresh routine run has `resume_id = None`, so no `ResumeInvalid` re-persist). The duplicate comes from `mergeFeedItem`: it only collapsed **consecutive** identical `user_message`s and exempted `user_message` from the dedup guard. So the engine's live WS **echo** of the turn — arriving after the assistant reply, on top of the hydrated history — was appended instead of recognized as a re-delivery. The trailing empty "Mission log" is the `deriveStatus` thinking indicator that shows whenever the feed ends in a `user_message`, which confirms the feed ended on a duplicate user turn.

**Fix:** `mergeFeedItem` now takes `opts.fromWs`. A WS-sourced `user_message` that duplicates one already in the feed is dropped (the engine emits each user message once per turn). Optimistic/local pushes omit the flag and always append, so a user **deliberately** re-sending the same text keeps both copies — only the WS handler (`use-session-events.ts`) sets `fromWs`, so no optimistic send path changes. This also stops the live store from ever accumulating a WS user duplicate regardless of hydrate-vs-echo ordering.

Files: `ui/chat/src/feed-merge.ts`, `app/src/stores/feeds.ts`, `app/src/hooks/use-session-events.ts`, tests in `ui/chat/test/feed-merge.test.mjs`.

Both commits are needed: `mergeFeedHistory` (assistant streaming/final) + `fromWs` (user-message re-delivery).

PR #479, opened 2026-06-09
fix(skills): make installed skills appear — YAML-safe frontmatter, preserve + feature, live Windows mirror (#398)

### LINKED ISSUE #398: Skill doesn't appear after being installed
- Even after relaunch the skill is not available.
- The skill folder is created in `~/.houston/workspaces/<ws>/<agent>/.claude/skills`

SO: win 11

cravenceiling:
Closes #398 ("Skill doesn't appear after being installed").

Three independent bugs kept installed skills from showing up. All are OS-agnostic except the last.

## 1. Frontmatter serialized to invalid YAML (the `ai-sdk` case) — most severe

A skill could install but **never appear in the list, even after restart**, while its folder sat in `.agents/skills`. The Vercel `ai-sdk` skill is the canonical case: its description contains `": "` (a YAML mapping indicator) and inner quotes. `format::serialize` emitted string scalars **unquoted**, producing invalid YAML that `parse_content` rejects — so `list_skills` **silently skipped it** (`Err(e) => tracing::warn!("skipping…")`). `ai-elements` (no colon in its description) parsed fine — exactly the difference reported.

**Fix:** emit every frontmatter string (name, description, category, image, tags, integrations, input fields, single-line prompt_template) as a proper YAML scalar — plain when unambiguously safe, otherwise double-quoted + escaped; flow-sequence items additionally guard `,` `[` `]` `{` `}`. Image URLs stay unquoted (no `": "`), so common output is unchanged. Verified end-to-end against the real `vercel/ai` `ai-sdk` SKILL.md — it now lists.

## 2. Install rebuilt a bare, non-featured skill (all OSs)

Install ran the fetched SKILL.md through `create_skill`, which rebuilt a bare frontmatter: `featured: false`, no image/category/integrations. The chat empty state shows only featured skills when any exist, so a freshly installed skill never appeared on the cards and lost its icon.

**Fix:** route both install paths through `houston_skills::install_skill_md`, which preserves the author's frontmatter and marks the skill `featured: true` (the documented default so a user finds what they installed). The install slug owns both the directory and the frontmatter `name`; over-long descriptions are clamped rather than failing validation.

## 3. Windows `.claude/skills` discovery mirror (Windows only)

The node that makes a skill natively discoverable to Claude Code was a privilege-gated symlink; on stock Windows 11 it failed (os error 1314), and the prior copy fallback goes stale.

**Fix:** use a Windows **directory junction** — no privilege required and *live*. Tier is symlink → junction → copy (last resort). Removal is junction-aware. `junction` crate added as a Windows-only dependency.

## Verification

- `cargo test -p houston-skills --features remote` — 70 green (new serialize + install tests; real `ai-sdk` round-trip verified manually).
- `cargo test -p houston-engine-core` — 332/333 green; the one failure (`login_relay::device_auth_relay_emits_url_then_code`) is pre-existing and unrelated (spawns OAuth child processes; confirmed it fails with these changes stashed and references no skills code).
- `cargo test -p houston-engine-server --test skills --test portable` — green.
- `cargo check --target x86_64-pc-windows-gnu -p houston-engine-core --tests` — Windows junction path compiles clean.

## Note for already-broken installs

A skill installed *before* this fix (e.g. a malformed `ai-sdk`) is already corrupt on disk and will keep being skipped; reinstalling reports "already installed". Recover by deleting its folder under `.agents/skills/<name>` and installing again.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2576,6 +2576,7 @@ dependencies = [
  "houston-skills",
  "houston-terminal-manager",
  "houston-ui-events",
+ "junction",
  "once_cell",
  "rand 0.8.5",
  "regex",
@@ -3346,6 +3347,16 @@ checksum = "5dea2b27dd239b2556ed7a25ba842fe47fd602e7fc7433c2a8d6106d4d9edd70"
 dependencies = [
  "serde",
  "serde_json",
+]
+
+[[package]]
+name = "junction"
+version = "1.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8cfc352a66ba903c23239ef51e809508b6fc2b0f90e3476ac7a9ff47e863ae95"
+dependencies = [
+ "scopeguard",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
--- a/engine/houston-engine-core/Cargo.toml
+++ b/engine/houston-engine-core/Cargo.toml
@@ -45,6 +45,10 @@ base64 = "0.22"
 # subprocess that popped its own progress dialog and blocked engine
 # boot.
 sevenz-rust2 = "0.18"
+# Directory junctions for the `.claude/skills/<name>` discovery mirror when a
+# real symlink is denied (stock Windows without Developer Mode). Junctions need
+# no privilege and stay live, unlike a copied snapshot (see src/skills.rs).
+junction = "1"
 
 [dev-dependencies]
 tempfile = "3"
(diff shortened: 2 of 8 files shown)


cravenceiling:
Follow-up commit `a7df771`: reinstalling now **heals a corrupt skill** automatically. A skill written before the YAML-quoting fix is unparseable on disk and was skipped by `list_skills`, yet the install guard still reported "already installed" — wedging the user. The guard now only blocks when the existing `SKILL.md` actually parses; a corrupt/empty one is replaced on reinstall. So an already-broken `ai-sdk` recovers by simply reinstalling it (no need to hand-delete the folder).

Issue #453, opened 2026-06-05
Hide claudecode connection dialog in the desktop app


cravenceiling:
Implement the same auto-detect logic the openai provider connection has for anthropic so the user doesn't see the anthropic connection dialog used in the webapp for the connection with the headless engine.

PR #511, opened 2026-06-11
fix(provider): supersede stale login on relaunch so retry isn't blocked (HOU-438)

cravenceiling:
## Linear

**HOU-438** — https://linear.app/houston-ai/issue/HOU-438/launch-provider-login-claude-sign-in-already-pending-blocks-retry

## Root cause

The engine keeps **one OAuth login slot per provider** (`login_relay::insert_session`). If a prior Claude/codex sign-in never finished or was cancelled, the slot stays occupied (up to the 10-minute relay timeout, or for the whole abandoned-tab window), so the next `launch_login` is rejected with `BadRequest("… sign-in is already pending")` — surfaced as the Sentry `HoustonEngineError`.

The frontend's intended recovery is to call `cancelLogin` before `launchLogin`, but it is applied **inconsistently across surfaces**, and the Cancel affordance is bound to ephemeral local React state that desyncs from the engine's session map:

- **In-chat reconnect (the surface token-expiry actually renders): `provider-reconnect-card.tsx`** calls `launchLogin` with **no** `cancelLogin` first, and its only buttons are "Sign in" / "Try Again" — both re-call the same cancel-less path. There is **no Cancel button** on this card, so an "already pending" rejection has no in-UI recovery (it loops on the same error). `auth-reconnect-banner.tsx` is worse — it swallows the error in an empty catch.
- **Settings / Picker** *do* show a Cancel button, but it is gated on local `pendingId`: the `handleConnect` catch clears `pendingId` the instant "already pending" fires (removing the button), and a fresh mount starts at `pendingId = null`. There is **no engine endpoint to resync in-flight login sessions**, so after an unmount/reopen the UI can't know a slot is held.
- Only `provider-error-cards/auth.tsx` does `cancelLogin → launchLogin`.

**Desktop vs remote.** On the desktop sidecar, an app restart kills the engine and wipes the in-memory `LOGIN_SESSIONS`, so the stale-slot window there is **one engine lifetime** (abandon a tab, retry within 600s). The durable case is **remote/persistent engines** (Always-On, container, mobile against a hosted engine) that outlive the client tab/app — closing and returning leaves the slot held. That is the primary HOU-438 scenario.

> Note: an adversarial audit corrected two claims from an earlier draft of this description — #453 hides only the `ProviderLoginUrl` dialog (not the card Cancel button), and desktop app-restart self-heals (sidecar dies). The corrected rationale is in the `mod.rs` comment.

## Fix

Make `launch_login` **idempotent** at the engine layer: it silently supersedes any stale pending session for that provider before spawning the fresh one (latest-launch-wins). This fixes **every** surface at once — including the cancel-less `provider-reconnect-card` / `auth-reconnect-banner` — instead of depending on each frontend to cancel-first.

- **Silent supersede** (`supersede_session` + a `superseded` flag): tears the stale subprocess down like a cancel, but emits **no** `ProviderLoginComplete` — a benign completion would otherwise race the replacement login and clear its spinner.
- **Orphan guard**: if `insert_session` still rejects (a genuinely concurrent Connect inside the 3s probe window), kill the just-spawned child instead of dropping it unkilled (`kill_on_drop` is `false`).
- **`biased;` relay select**: drain buffered stdout before observing the child's exit, so a fast-exiting CLI no longer drops `ProviderLoginUrl`. (Convergent with HOU-446, which added the same `biased;` on `main` — reconciled in the merge.)

## Files changed

- `engine/houston-engine-core/src/provider/login_relay.rs` — `supersede_session`, the `superseded` flag, the silent `Superseded` outcome, `biased;` select, 2 new tests.
- `engine/houston-engine-core/src/provider/mod.rs` — `launch_login` pre-supersede + insert-reject orphan kill.

No frontend changes. (Follow-up worth considering, but not required by this fix: add cancel-before-retry to `provider-reconnect-card` and stop the silent catch in `auth-reconnect-banner`.)

## Tests

- `supersede_session_frees_slot_without_emitting` — supersede frees the slot **and** the relay stays silent.
- `supersede_unblocks_a_fresh_insert` — regression: after supersede, a fresh insert succeeds instead of "already pending".
- Full `houston-engine-core`: **370 passed**; relay tests 18/18 across repeated runs (no flake).

## Reproduce the bug BEFORE this fix

Easiest on a persistent/remote engine (or within the 600s window on desktop):

1. Start a Claude sign-in but do **not** complete the browser flow:
   `curl -X POST http://127.0.0.1:<port>/v1/providers/anthropic/login`
2. Without cancelling, start it again:
   `curl -X POST http://127.0.0.1:<port>/v1/providers/anthropic/login`
3. **Before:** the second call returns `400 … claude sign-in is already pending`. In-app, a chat whose provider needs re-auth renders `provider-reconnect-card`, whose "Sign in" / "Try Again" both hit this rejection with no Cancel button to free the slot — stuck until the 600s timeout or (desktop) an app restart.

## Verify the fix AFTER

1. Same steps: `POST …/login`, abandon, then `POST …/login` again.
2. **After:** the second call returns `200` — the stale subprocess is killed and a fresh sign-in spawns. Connect / Sign in / Try Again is retryable immediately from **any** surface (Settings, picker, onboarding, in-chat reconnect card), desktop and remote, with no restart.
3. The superseded teardown emits no `ProviderLoginComplete`, so the new attempt's spinner is not cleared.

### DIFF
--- a/engine/houston-engine-core/src/provider/login_relay.rs
+++ b/engine/houston-engine-core/src/provider/login_relay.rs
@@ -35,7 +35,7 @@ use houston_ui_events::{DynEventSink, HoustonEvent};
 use once_cell::sync::Lazy;
 use regex::Regex;
 use std::collections::HashMap;
-use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::io::{AsyncBufReadExt, AsyncReadExt, AsyncWriteExt, BufReader};
@@ -76,15 +76,23 @@ struct LoginSession {
     /// Fired by [`cancel_login`] to abort an in-flight browser sign-in.
     /// The relay task holds its own `Arc` clone and selects on it.
     cancel: Arc<Notify>,
+    /// Set by [`supersede_session`] immediately before it fires `cancel`,
+    /// marking this teardown as "a fresh launch is replacing me." The relay
+    /// reads it on the cancel path and stays silent (no
+    /// `ProviderLoginComplete`) so it can't clear the replacement login's
+    /// spinner on the frontend (HOU-438). A user-driven [`cancel_login`]
+    /// leaves it false, preserving the benign completion that re-arms the card.
+    superseded: Arc<AtomicBool>,
     /// Identifies which relay owns this map entry (see [`SESSION_SEQ`]).
     token: u64,
 }
 
 /// Handed back by [`insert_session`] to the spawn site so the relay
-/// task gets the same cancel handle + token stored in the map.
+/// task gets the same cancel handle + supersede flag + token stored in the map.
 #[derive(Debug)]
 pub(super) struct RelayRegistration {
     cancel: Arc<Notify>,
+    superseded: Arc<AtomicBool>,
     token: u64,
 }
 
@@ -94,6 +102,10 @@ pub(super) struct RelayRegistration {
 enum RelayOutcome {
     Exited(std::io::Result<std::process::ExitStatus>),
     Cancelled,
+    /// A fresh launch tore this session down to take its slot. Behaves like
+    /// `Cancelled` (kill + reap the child) but the relay emits no
+    /// `ProviderLoginComplete` — see [`supersede_session`].
+    Superseded,
 }
 
 /// Regex over a single line of CLI stdout, looking for an HTTPS URL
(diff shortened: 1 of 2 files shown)


cravenceiling:
Closing unmerged. HOU-438's observed trigger was already fixed in v0.4.22 by #490 (cancel-first reconnect card), which frees the engine's stale login slot before relaunching — last Sentry event was 0.4.21 (2026-06-10), zero since 0.4.22.

This PR was engine-side defense-in-depth (auto-supersede in `launch_login`) covering the remaining cancel-less surfaces (Settings/Picker, the secondary ProviderReconnectCard, onboarding, Gemini dialog) plus remote/persistent engines that #490 doesn't reach. Real but not worth carrying now. Branch `claude-signin-pending-blocks-retry` is preserved if we want the remote-engine hardening later.

Linear HOU-438 → Done; Sentry HOUSTON-APP-62 → resolved.

PR #521, opened 2026-06-12
fix(sentry): deduplicate Sentry issues via stable fingerprints (HOU-449)

cravenceiling:
## Root cause

The `houston-app` Sentry project was drowning in **duplicate issues** because event messages embed volatile data, so Sentry's default `name`+`value` grouping spawned a *new issue per variant*. Pulled the live issue list (14d) to confirm — two families dominated:

| Family | Example titles | Why they split |
|---|---|---|
| **A — renderer transport drops** | `read_agent_file: Failed to fetch (127.0.0.1:56736)`, `… (127.0.0.1:61449)`, `check_provider_status: Load failed (127.0.0.1:49599)` | the local sidecar's port is **random per restart** → ~40 issues that are really a couple of transport drops |
| **B — engine Codex parser** | `Failed to parse Codex event: invalid type: null, expected i32 at line 1 column 190` … `column 300` … (plus `expected value at line 1 column 1` as 5+ separate issues) | the serde **column moves every malformed frame**, AND the raw frame is appended on a second line → ~50 issues for one parser bug |

Proof the approach works: `houston-engine subprocess exited unexpectedly …` is already ONE issue because `engine_supervisor.rs` sets an explicit `engine-subprocess-exit` fingerprint. This PR generalizes that.

## Fix

Both Sentry clients now stamp a deterministic `fingerprint` in their `before_send` (Rust) / `beforeSend` (JS) hook, computed from the **first line** of the message with volatile tokens masked (IPv4`:port`, UUIDs, `0x` hex, bare digit runs → `{addr}`/`{uuid}`/`{hex}`/`{n}`). Each family collapses into ONE issue.

- It only overrides Sentry's **default** fingerprint, so deliberate ones (the supervisor's `engine-subprocess-exit`) still win.
- It **no-ops on an empty message** (keeps default grouping — never collapses unrelated empty events into one bucket).
- Never drops an event.

The **app-process Rust** client is intentionally left as-is: its messages are constant (updater URL, WebView2 HRESULT, panics) and already land as one issue each. Documented as a deliberate scope decision.

### Files
- `app/src/lib/sentry-fingerprint.ts` (new) — renderer normalizer + `fingerprintForEvent`, wired into `app/src/lib/sentry.ts` `beforeSend`.
- `engine/houston-engine-server/src/sentry_grouping.rs` (new) — engine normalizer + `with_stable_fingerprint`, composed **after** the existing secret scrub in `main.rs` `before_send`. Adds `regex` dep.
- `knowledge-base/production-infra.md` — documents the grouping fix + the resolve→regression investigation below.
- Tests: `app/tests/sentry-fingerprint.test.ts` (9) + in-module Rust tests (6), all using verbatim Sentry titles.

## Second checkbox — "does a resolved issue reopen when it recurs?"

Yes, and it keys **entirely off the fingerprint**:
- An event matching a **resolved** issue's fingerprint flips it back to *Unresolved* and records a **regression** (Activity tab + regression alert if configured).
- Plain *Resolve* reopens on any later match; *Resolve in next release* stays resolved until a match lands in a release semver-greater than the one current at resolve time; *Archive/Ignore* mutes and auto-unarchives only on escalation.
- **The catch (why this shared a ticket with dedup):** an *unstable* fingerprint breaks recurrence tracking — a recurrence arriving with a different port/column becomes a brand-new issue instead of reopening the resolved one. **So the stable fingerprints in this PR are the prerequisite that makes resolve→regression a trustworthy "did the fix hold?" signal.**

## Verification

**Before (reproduce the dup sprawl):**
1. Open the Sentry project `houston-cd/houston-app` → Issues, `statsPeriod=14d`, sort by frequency.
2. Observe dozens of near-identical issues differing only by `(127.0.0.1:<port>)` or `column <N>`.

**After (unit-proven now; observable once a build ships):**
- `cd engine && cargo test -p houston-engine-server --bin houston-engine sentry_grouping` → 6 pass (different columns/ports share one key; explicit fingerprints preserved; empty message keeps default).
- `cd app && node --experimental-strip-types --test tests/sentry-fingerprint.test.ts` → 9 pass; `pnpm tsc --noEmit` clean; full suite 293/293.
- Logic check: `normalizeFingerprintMessage("Failed to fetch (127.0.0.1:56736)") === normalizeFingerprintMessage("Failed to fetch (127.0.0.1:61449)")` → both `"Failed to fetch ({addr})"`.
- On a DSN-baked build, repeated transport drops / Codex parse failures land in a single grouped issue instead of one-per-port/column.

**Existing backlog:** this only stops *new* dupes. Merge the historical duplicates by hand in the Sentry UI (Issues → multi-select → **Merge** — irreversible, so deliberately not scripted).


cravenceiling:
Closing in favor of server-side Sentry Fingerprint Rules (Project Settings → Issue Grouping) instead of shipping grouping logic in the app/engine code. Rationale: the codebase is expected to migrate (desktop → webapp), and server-side rules are decoupled from the code, so they survive a rewrite as long as the Sentry project is kept. The one-time backlog cleanup (4,791 → ~100 unresolved, duplicates resolved) already happened and is unaffected by closing this.

PR #605, opened 2026-07-03
fix(credentials): one OpenCode key connects both Zen and Go gateways

cravenceiling:
## What & why

Connecting **OpenCode** with an API key in the cloud left **OpenCode Go** unusable: sending a first message on a Go model showed a **"Sign in to OpenCode Go again"** card even though the key was saved.

Root cause (from prod logs + the code): `opencode` (Zen) and `opencode-go` (Go) are two pi gateways — `opencode.ai/zen/v1` vs `opencode.ai/zen/go/v1` — but they authenticate with the **same opencode.ai key** (pi reads `OPENCODE_API_KEY` for both). Houston's cloud/web path treated them as independent credentials, so the saved key only credentialed `opencode`; a Go turn hit `opencode-go`'s empty credential and surfaced the reconnect card.

The desktop Tauri adapter already fans the key across both gateway ids (HOU-577, `synthetic.ts`/`client.ts`), but **the cloud/web path has no such adapter** — so nothing shared the key there. This PR enforces the sharing **backend-side**, so one saved key connects both gateways regardless of client.

## Change

- **`@houston/domain` `credentialSiblings`** — the single source of the Zen↔Go pairing (byte-mirrored in the runtime at `packages/runtime/src/auth/credential-siblings.ts`, which can't import `@houston/domain`, the same reason domain mirrors `ProviderId`).
- **host `resolveSharedCredential`** — the sandbox serve endpoint (`/sandbox/credential`) and the per-turn credential fetch (`freshCredential`) fall back to a sibling's stored credential, **relabeled** to the requested gateway.
- **runtime** — `applyServedCredential` / `setApiKey` / `logout` now write and clear **both** gateways' `auth.json` entries together, so pi's `has()` (connected status) and `getApiKey()` (the turn) both resolve the shared key.

Composes with the existing desktop fan-out (idempotent); makes it unnecessary but harmless.

## Tests

New/extended, all green:
- `credentialSiblings` mapping — domain + runtime
- `applyServedCredential` mirrors Zen↔Go, and does **not** mirror a provider with no sibling
- `resolveSharedCredential` — own-credential, sibling fallback + relabel, null-when-neither, no cross-wiring
- serve endpoint serves a Go turn the shared key via sibling fallback

`domain` + `runtime` + `host` suites pass (706 tests); `typecheck` (all three) and `pnpm check:boundaries` clean; Biome clean.

## Verification note

Logic is covered by unit/integration tests. End-to-end confirmation (paste an OpenCode key → run an OpenCode Go model like `kimi-k2.6` and get a normal reply, no reconnect card) needs this built into the cloud engine-pod image + rolled out — do that after merge.

### DIFF
--- a/knowledge-base/agent-manifest.md
+++ b/knowledge-base/agent-manifest.md
@@ -270,6 +270,15 @@ Notes:
   directly) — the model picked selects the gateway; opencode.ai enforces
   Go-subscription vs Zen-credit entitlement per request (surfaced as a
   provider-error card), so Houston never has to detect the plan from the key.
+  The **cloud/web path has no Tauri adapter**, so the client-side fan-out above
+  doesn't run there — the same sharing is enforced BACKEND-side instead:
+  `credentialSiblings` (`@houston/domain`, byte-mirrored in the runtime at
+  `packages/runtime/src/auth/credential-siblings.ts`) makes the host serve either
+  gateway from the one stored opencode key (`resolveSharedCredential`) and the
+  runtime write/clear both gateways' `auth.json` entries together
+  (`applyServedCredential`, `setApiKey`, `logout`). So a single saved key connects
+  both gateways with or without a client fan-out; without it a Go turn hit
+  `opencode-go`'s empty credential and surfaced a spurious "sign in to OpenCode Go".
 - _[NEW ENGINE only]_ The TS engine also adds **GitHub Copilot**
   (`github-copilot`) as a **subscription OAuth** provider — pi-ai ships it
   built-in (no adapter, no CLI), so it's registry entries only across the same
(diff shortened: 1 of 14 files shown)


cravenceiling:
Superseded by #606. This PR chased a credential-sharing theory that turned out **not** to be the cause: live probing of the agent pod showed the OpenCode key was reaching both gateways fine (`keysEqual:true`) and authenticating (`GET /zen/v1/models` → 200). The actual bug is that opencode.ai returns HTTP 401 for non-auth failures (insufficient balance, unsupported model) and Houston's classifier mapped every 401 to a reconnect card. Fixed correctly in #606.

PR #664, opened 2026-07-06
feat(app): show a loading state while a new agent's engine warms up (HOU-693)

cravenceiling:
Fixes HOU-693.

## Problem

On the hosted profile, creating an agent answers instantly while its engine warms up in the background for a couple of minutes (HOU-649), and the platform exposes no readiness field, status endpoint, or per-agent ready event. The agent appears in the sidebar immediately, so:

- every tab (routines, files, settings…) sat on its own endless "Loading..." spinner with no explanation, and
- a first chat message never left the composer: creating the conversation awaited the activity write, which is held server-side for the whole cold start.

What the platform does guarantee: any per-agent request is held until the engine is reachable, then answered in arrival order.

## Fix (client-side + one additive host change)

**Provisioning state** (`app/src/stores/agent-provisioning.ts`): `useAgentStore.create` marks a just-created agent (hosted mode only, no-op co-located); a readiness long-poll (a cheap per-agent read, held server-side until the engine answers) clears it. Any definitive HTTP answer counts as "reachable"; 502/503/504 and transport failures mean "keep waiting". 10-minute cap with an error toast (the deadline races the in-flight read, so a wedged request can't postpone it). Survives relaunch via a TTL-bounded localStorage mirror. Rename mid-warm-up re-keys the entry; delete clears it. Pure logic in `app/src/lib/agent-provisioning.ts`, covered by node tests.

**Tabs** (`AgentProvisioningPlaceholder`): while provisioning, every non-activity tab renders a centered "We are creating your agent — This can take a couple of minutes." placeholder in the tab body itself (where the stuck spinner used to be). The tab's queries don't mount until the engine is ready.

**Chat — optimistic first send**: `POST /agents/:id/activities` now honors an optional client-generated `id` (protocol + host route + both adapters; validated; upsert keeps a same-id retry idempotent). `createMission` gains a warming path: generate the id client-side, queue the board-row write first (arrival order keeps it ahead of the turn's board-status write), start the turn immediately — the SDK pushes the user's bubble into the feed synchronously while the wire send is held — and return at once. The composer clears instantly, the `AgentProvisioningCard` ("It will reply to your message as soon as it is ready.") sits under the sent message, and when the engine wakes the reply streams in on its own and the mission card lands on the board. The card unmounts the moment the readiness probe clears.

i18n en/es/pt (`shell:agentProvisioning.*`, `chat:errors.missionRowFailed`); design inventory v3 adds `agent-provisioning-card` (chat-card + tab-placeholder variants, app/-locked `partial` on web).

## Verification (performed end-to-end)

Simulated the hosted cold start locally: hermetic TS host behind a proxy that holds all per-agent routes for 90s after agent creation, web UI pointed at it as a remote engine, driven with Playwright.

- Create agent → Routines tab shows the centered placeholder (screenshot-verified), not a stuck spinner.
- New chat → send → the message appears as a sent bubble immediately, composer clears, the provisioning card renders under it.
- After the cold window: the card removes itself with no user action, the held send is processed (turn ran; with no valid provider it surfaced the normal reconnect card — with a real provider the reply streams), and the mission card appears on the board exactly once with the client-generated id. No degraded "Couldn't update the board status" messages.
- Desktop/local (co-located engine): nothing is ever marked — behavior unchanged.
- Gates: Biome, app + web typecheck, `check-locales`, `check:parity`, 579 app tests, 550 host tests (including new coverage for the client-supplied activity id).

## Notes

- The probe logic is deliberately app/-side but written SDK-shaped (pure, dependency-injected) for extraction when mobile grows a create-agent flow.
- A 500 from the agent's engine clears the "creating" state on purpose: something answered, and the user's own requests surface the real error.


cravenceiling:
## Update: warm-up sends are now a client-side queue (real-cloud feedback)

Testing against the deployed cloud surfaced the flaw in the first design: the "held send" lived in the browser's network stack, so infrastructure timeouts killed it mid-warm-up (error card, no reply ever) and a relaunch aborted it and wiped the in-memory bubble (message disappeared, bare "Mission in progress" at the top).

**New design** (`app/src/lib/warming-sends.ts`): a message to a still-warming agent renders as a local bubble immediately and queues on the agent's provisioning entry (persisted, TTL-bounded, reload-safe). The readiness probe's `onReady` flushes the queue: the board row is upserted first (client-generated id — a retry of an already-landed row is a no-op), then each send fires with `suppressUserBubble`. No wire request is held during the warm-up at all, so there is nothing for load balancers or reloads to kill. Follow-up sends into a warming agent take the same queue.

Three adjacent boot bugs this flow surfaced, all fixed:
- `activeOld()` (synthetic-workspace provider label) sat on the boot path and is a per-agent read — held for the whole warm-up, wedging the "Loading your workspace" splash after a relaunch. Now bounded (4s → defaults), per the repo's boot-gate invariant.
- `useHoustonInit` awaited per-agent scheduler starts before restoring the last-used agent — same held-request wedge. No longer awaited.
- `useHoustonInit` read `last_agent_id` AFTER `loadAgents`, whose auto-selection side effects overwrite that preference — so a relaunch always restored `agents[0]`, never the actual last agent. Generic pre-existing bug; the preference is now read first.

Copy: placeholder body is now "This may take a moment." (en/es/pt).

**Re-verified end-to-end** against the simulated cold engine, including a relaunch mid-warm-up: bubble + card at send; boot lands back on the right agent instantly after reload; the message and its board row survive; the queued send fires exactly once on ready; the turn runs (typed provider card with the harness's fake key — a real provider streams the reply).

cravenceiling:
## Update: free exploration during warm-up + card-only chat wait

Rebased onto latest main. UX rework per feedback:

- **Chat**: while the agent is being created, a sent message shows the bubble + the provisioning card only — the "Mission in progress" indicator no longer competes with the card (`deriveStatus` gains a props-only `suppressPendingIndicator`, threaded ui/chat → ui/board → both boards). When the agent is ready the card removes itself, the queued message is delivered, and the normal turn indicator/reply take over.
- **Tabs unblocked**: the full-tab placeholder is gone. Per-agent reads answer instantly with "nothing yet" during warm-up (a fresh agent has no data; ready-time events refetch the real state), so routines/skills/files/settings render their true empty states and the user can explore freely.
- **Writes gated by a dialog**: any save/create/update/upload/rename/run-now on a warming agent opens a centered "We are creating your agent — It is almost ready…" dialog with the boot splash's helmet-in-motion, then rejects with a typed error that is never toasted (wrapper surface bypassed + a global unhandled-rejection exemption, mirroring the benign Web-Locks pattern). The dialog dismisses itself if the agent finishes while open. The app's own post-create setup writes (provider/model config, AI-routine seed) keep their held-request behavior via an explicit `allowWhileWarming` opt-out.
- Copy: "This may take a moment." everywhere; new `shell:agentProvisioning.blockedBody` (en/es/pt).

Verified end-to-end on the simulated cold engine: tabs explorable with empty states, routine save → dialog, chat = bubble + card with no shimmer, delivery on ready. Gates: Biome, typechecks, locales, parity, 751 app + 621 host tests.

PR #752, opened 2026-07-07
feat: inline waveform recording UX for dictation

ja-818:
## What

Replaces the small pulsing-dot recording indicator (#745) with the reference-quality inline pattern: **recording takes over the composer pill** — attach button stays at the left, the textarea becomes a live waveform track (thin rounded amplitude bars filling left-to-right, dotted leader for the remaining time; 30 s fills the track, then it compresses), with **✕ cancel** and a **filled ✓ accept** at the right. Requesting shows an all-dots pulse; transcribing freezes the waveform dimmed under a scanning shimmer with a spinner until the transcript lands in the input.

## How

- Recorder computes per-100 ms RMS levels (sqrt curve, 1200-bucket cap) in a DOM-free `LevelAccumulator`; `DictationControl` gains `getLevels?: () => readonly number[]` (0..1), polled with rAF only while capturing — no re-render storms.
- Waveform renders on a DPR-scaled canvas using `currentColor` from a `text-muted-foreground` wrapper (theme-safe, no hardcoded hex); track layout + peak-preserving downsampling live in a pure tested module.
- Escape-cancel moved to a document listener while capturing (the textarea is unmounted); submit swaps for cancel/accept; `role=status` announcements kept; nothing hover-gated.

## Verification

117 dictation tests (20 app + 97 ui/chat), full app suite 959, repo typecheck/biome/locales clean. Recording and transcribing states screenshot-verified against the reference design.

### DIFF
--- a/app/src/lib/dictation/levels.ts
+++ b/app/src/lib/dictation/levels.ts
@@ -1,0 +1,71 @@
+/**
+ * Pure amplitude-metering helpers for the dictation waveform. No DOM, no Tauri
+ * — the recorder (`recorder.ts`) feeds raw Float32 frames in, and this module
+ * turns them into a compact per-100ms level history the composer renders as a
+ * live waveform. Kept separate so the bucketing/normalization math runs under
+ * the bare Node test runner (`app/tests/dictation-levels.test.ts`).
+ */
+
+/** One amplitude bucket per this many ms of captured audio. */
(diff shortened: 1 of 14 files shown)


ja-818:
Web-check failures here were inherited from main (#785 regressions, fixed in #791). Re-running CI on a rebased head after #791 lands.

PR #766, opened 2026-07-07
fix(chat): mission log stays open and shows thinking + tool calls everywhere (HOU-717)

cravenceiling:
Fixes HOU-717 — the "Mission in progress" log was unusable: the dropdown snapped shut on its own while the agent worked, and most of the time it showed nothing (no thinking, no tool calls).

## Root causes and fixes

**1. The open log closed by itself (the "buggy dropdown").**
The process block's React key was `process-<first>-<last>` over its segments. A streaming turn keeps appending segments, so the key changed on every new thinking/tool block, React remounted the block, and the Collapsible's `isOpen` state reset to closed. The key is now derived from the first segment only, which is stable for the life of the block (`ui/chat/src/chat-process-groups.ts`).

**2. A reopened conversation's mission log was empty.**
Persisted history (`ChatMessage`) carried no reasoning at all and only bare tool names (no inputs). Now:
- `ChatMessage.thinking` and `ToolCallRecord.input` are persisted (additive protocol fields),
- the runtime records them from the turn's wire frames (`exec-turn.ts`),
- the SDK history fold replays them — thinking before tools, matching live render order (`history.ts`),
- a settle-from-history adopts the persisted reasoning when nothing streamed live.

**3. Opening a RUNNING mission showed nothing until the next frame.**
A fresh SSE attach gets a `sync` frame that only carried the partial assistant text. The shared snapshot reducer now accumulates the running turn's thinking and tool calls, the `sync` frame carries them (additive, legacy servers/clients unaffected), and the turn sink replays them into the feed, deduped across resyncs so nothing doubles (`snapshot.ts`, `turn-sink.ts`). The host relay's snapshot parser carries the new fields through (no bus dialect bump needed — the fields are additive in both directions).

## Reproduction (before the fix)

1. Ask an agent something that runs several tools (e.g. "list my workspace files and summarize each one").
2. While "Mission in progress: …" is showing, click the chevron to open the log → within a second or two (as the next thinking/tool block streams in) the pane closes by itself.
3. Let the mission finish, close and reopen the conversation → open "Mission log": no reasoning at all, tools show names only.
4. Start a long mission, close the conversation, reopen it while the agent is still working → the log shows nothing that happened before you reopened.

## Verification (after the fix)

1. Same mission: open the log while the agent works → it stays open, thinking and tool calls stream in live.
2. Reopen a settled conversation → "Mission log" replays the reasoning and each tool call with its input.
3. Reopen a conversation mid-run → the log shows the thinking and tool calls that streamed before you attached, and continues live without duplicates.
4. `pnpm --filter @houston/protocol --filter @houston/runtime-client --filter @houston/runtime --filter @houston/sdk --filter @houston/host --filter @houston/domain test`, `pnpm -r typecheck`, `pnpm check`, `pnpm check:boundaries`, and `cd ui/chat && pnpm test` all pass. New tests cover the stable process key, the history fold, the snapshot accumulation, the sync replay dedup, and the runtime persistence.

### DIFF
--- a/packages/host/src/turn/relay-dialect.ts
+++ b/packages/host/src/turn/relay-dialect.ts
@@ -55,6 +55,20 @@ export function parseSnapshot(raw: string | null): ConversationSnapshot {
       partial: s.partial,
       seq: s.seq,
       ...(s.turnId ? { turnId: s.turnId } : {}),
+      // The running turn's activity (HOU-717) — ADDITIVE fields, so no
+      // dialect bump: an old replica's parse simply drops them (the sync
+      // degrades to text-only), and an old snapshot reads as "no activity".
+      ...(typeof s.thinking === "string" ? { thinking: s.thinking } : {}),
+      ...(Array.isArray(s.tools)
+        ? {
+            tools: s.tools.filter(
+              (t): t is NonNullable<ConversationSnapshot["tools"]>[number] =>
+                t !== null &&
+                typeof t === "object" &&
+                typeof (t as { name?: unknown }).name === "string",
+            ),
+          }
+        : {}),
     };
   }
   return EMPTY_SNAPSHOT;
--- a/packages/protocol/src/conversation.ts
+++ b/packages/protocol/src/conversation.ts
@@ -150,6 +150,13 @@ export type ChatRole = "user" | "assistant";
 
 export interface ToolCallRecord {
   name: string;
+  /**
+   * The tool call's arguments, exactly as the live `tool_start` frame carried
+   * them. Persisted so a reloaded conversation's mission log shows WHAT each
+   * tool did (the command run, the file written), not just the tool's name
+   * (HOU-717). Absent on records written before this field existed.
+   */
+  input?: unknown;
   isError?: boolean;
 }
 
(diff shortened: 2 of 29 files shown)


cravenceiling:
Round 2, after user testing surfaced three more gaps (screenshot showed empty tool boxes and rows closing on their own):

1. **Tool output now flows end to end.** `tool_end` previously carried only `{name, isError}`, so an expanded tool row was always an empty box. Both backends (pi + Claude SDK) now extract the result text, clip it at the source (`TOOL_RESULT_PREVIEW_MAX`), and it rides the live feed, the mid-turn sync snapshot, and the persisted `ToolCallRecord.result` for reload replay.

2. **ui/chat understands the pi tool vocabulary.** Labels, icons, header details, and content renderers previously only matched Claude-style names (`Bash`, `Read`); pi's lowercase tools (`bash`, `read`, `write`, `edit`, `grep`, `find`, `ls`) with their own arg shapes (`path`, `edits[].oldText/newText`) all fell into a generic renderer that showed nothing. Both dialects now map to the same renderers, and an opened row never renders an empty box (it falls back to the call's arguments).

3. **Manually opened tool rows stay open.** `ToolBlock` auto-closed any open row 800ms after a result existed, including a settled row the user had just expanded. Auto-close is now armed only by an automatic open.

All suites green: runtime 547, sdk 259, runtime-client 61, host 636, ui/chat 88; full typecheck + Biome clean.

PR #779, opened 2026-07-08
feat: always show the chat beside the routine (HOU-725)

cravenceiling:
## What changed

**One surface, no chooser.** "New routine" no longer asks *chat vs form*: it opens the empty form on the left and starts the guided setup chat on the right, immediately. The agent speaks first, mentions it can help create the routine now **and** change it any time later from this same chat, then runs the existing one-question-per-turn interview.

**The chat is permanent and per-routine.** Routines gain an optional `setup_activity_id` (protocol + `routines.schema.json` + domain) linking them to their setup chat:
- Agent-created routines: the create kickoff carries the chat's own activity id and instructs the agent to write it when saving.
- Form-created routines: the client stamps the open draft chat's id on submit, so the two paths converge.
- Pre-existing routines with no chat: opening one starts a chat with a short modify greeting (no interview) and stamps the link.

Reopening a routine always resumes that same conversation beside the form. Setup chats no longer self-archive when done, the editor form live-syncs when the agent modifies the open routine from the chat (unless the user has unsaved edits), and after a chat-driven creation the empty creation form auto-switches to editing the new routine. Portable exports strip `setup_activity_id` (a chat id is machine-local).

**`description` removed.** Nothing functional consumed the routine description — the fire path uses only `prompt`, the grid renders only `name`. Removed from the form, types (protocol / engine-client / ui), schema, prompts, portable/anonymize/scan paths, and locales; a stray on-disk `description` is dropped on read exactly like the legacy `timezone` cleanup (idempotent, no migration).

**Test infra.** The fake host gets a real in-memory routines store (files-first, backed by `.houston/routines/routines.json`) so Playwright can exercise the create → link → reopen loop end to end.

## Before (reproduce the old behavior)

1. Open the Routines tab → **New routine** → a dialog forces a choice between "Create it in chat" and "Fill in the details myself".
2. Pick chat, finish the interview: once the chat settles and the panel closes, the conversation archives itself — reopening the created routine shows only the form, the chat is gone for good.
3. The editor shows a Description field that nothing anywhere reads.

## After (verify)

1. Routines tab → **New routine**: no dialog; empty form left, chat right, the agent greets first and says you can come back to this chat later.
2. Let the agent create the routine (or fill the form and hit *Create routine* while the chat is open) → the routine's entry in `.houston/routines/routines.json` carries `setup_activity_id`.
3. Go elsewhere, come back, open the routine → the same conversation, with its full history, is beside the form; continue talking to modify the routine and watch the form refresh.
4. Open a routine created before this change → a chat appears with a one-line "I can change this routine any time" greeting and stays attached from then on.
5. No Description field in the editor; the chat never appears as a board card; an unfinished creation still leaves the Continue/Discard banner on the grid.

## Gates

- `pnpm typecheck` (all 22 projects), `pnpm check` (Biome), `pnpm check:boundaries`, `pnpm check-locales` — clean
- `@houston/{host,runtime,domain,fake-host}` vitest: 637 + 554 + 96 + 6 passing
- `app` node tests: 1005 passing (prompt-contract tests updated for the persistence beats + new modify kickoff)
- Playwright full suite: 41 passing, incl. 4 rewritten/new routine-chat scenarios (form+chat open together; mid-interview banner; created routine claims the chat and resumes on reopen; legacy routine gets a linked chat on first open)


cravenceiling:
Pushed two follow-up fixes for the UI bugs found while testing (23052f0):

**Chat disappeared on window resize / fullscreen.** AIBoard closes its detail panel on any pointerdown outside the board or panel — which includes the titlebar interactions that resize/maximize the window and every other piece of app chrome. The routine chat is a persistent companion of the routine view, not a dismissible overlay, so AIBoard now takes `disableOutsideClose` and the setup chat sets it. The X button, tab switches, and unmount still close it.

**Chat closed "out of nowhere" when the agent modified the routine.** The chat↔routine link lived only in `routines.json` (`setup_activity_id`), which the agent rewrites when saving a change — a save that dropped the field unlinked the open chat at the exact moment the form showed the edit. The link is now bidirectional: the activity's `routine_id` (durable, agents never rewrite `activity.json`) is stamped on every claim, lookup prefers it, and a reconciliation effect restores a dropped forward link on the next refetch. The modify kickoff additionally instructs the agent to change only the fields the user asked about.

Coverage: pure resolver/heal functions with unit tests (1008 app tests green), plus two new Playwright regressions — chrome clicks never dismiss the chat, and a simulated agent link-drop self-heals with no duplicate chat after reload (43 e2e green).

cravenceiling:
Two more improvements pushed (c8952cc):

**Persistent chat, guaranteed on screen.** The conversation was already stored per routine and resumable, but the panel could still be closed out from under you. Now, while a routine is open, its chat is a fixed part of the view: no close button (the grid's draft chat keeps its X, with the banner as the way back), outside clicks already ignored, and a keep-open effect re-opens the panel after any stray close (tab round-trips, another surface stealing the shared panel container). The e2e for the claimed routine now also *continues* the resumed conversation — a follow-up message lands in the same session and gets a reply in the same thread.

**Agent knows which providers are connected.** Asking a routine to run on an unconnected provider ("use deepseek") used to get pinned anyway and could only fail at fire time. Both kickoff prompts now enumerate the user's actually-connected providers (from the shared provider-status query), instruct the agent to only ever pin one of those ids, to tell the user when a requested provider isn't connected (pointing at model settings), and never to invent provider/model names. While statuses are still loading the prompt stays generic rather than claiming nothing is connected.

Gates: 1009 app tests, 43 Playwright e2e, typecheck + Biome clean.

PR #849, opened 2026-07-11
feat(app): metadata-driven cloud-migration gate — user_metadata.migrated (HOU-719)

cravenceiling:
## What

The first-run cloud-migration gate now keys off a single authoritative, cross-machine boolean on the user's Supabase metadata — `user_metadata.migrated` — instead of inferring "is this an existing user" from local disk.

| `migrated` | Meaning | First-run surface |
|---|---|---|
| `true` | Done: migrated, or a new user past onboarding | App |
| `false` | Existing local-app user marked for migration | **Migration wizard** (only where a `~/.houston` actually exists on this machine) |
| absent | Brand-new cloud user | Normal onboarding |

The whole existing user base is backfilled to `false` at cutover (`scripts/backfill-migrated-flag.mjs`, idempotent, reconciles already-completed accounts to `true`), so **"absent = new"** reliably holds.

## Why

The previous gate inferred existing-vs-new from a local `~/.houston` scan plus a `migration_status` string. Keying the decision off server-side metadata makes it deterministic and cross-machine: a marked user is offered the wizard on whichever machine still holds their data, a migrated user never sees it again anywhere, and a truly new user goes straight to onboarding.

Local detection stays, but only in its correct role now: *does THIS machine hold the data to migrate* (a `false` user on a fresh machine keeps `false` and migrates on the device that has it).

## Changes
- `lib/migration-status.ts`: add `readMigrated` / `writeMigrated`.
- `hooks/cloud-migration-trigger.ts`: gate on `migrated` (only explicit `false` is a candidate; `true`/absent pass).
- `hooks/use-cloud-migration.ts`: feed `migrated`; only probe disk for marked users.
- `stores/cloud-migration-finish.ts`: stamp `migrated:true` on completion.
- `components/settings/sections/migration.tsx`: "Continue migration" gates on `migrated !== true`.
- `scripts/backfill-migrated-flag.mjs`: one-time cutover backfill (dry-run by default).

## Tests
- `cloud-migration-trigger.test.ts` updated + new "brand-new user (absent) → onboarding" case. 10 pass.
- `pnpm tsgo`, `pnpm check-locales`, `pnpm check` all green.

HOU-719

### DIFF
--- a/app/src/components/settings/sections/migration.tsx
+++ b/app/src/components/settings/sections/migration.tsx
@@ -4,7 +4,7 @@ import { useTranslation } from "react-i18next";
 import { useSession } from "../../../hooks/use-session";
 import { isHostedGatewayEngine } from "../../../lib/engine";
 import { reportError } from "../../../lib/error-toast";
-import { readMigrationStatus } from "../../../lib/migration-status";
+import { readMigrated } from "../../../lib/migration-status";
 import { osDetectLegacyHouston, osIsTauri } from "../../../lib/os-bridge";
 import { queryKeys } from "../../../lib/query-keys";
 
@@ -21,9 +21,7 @@ const STORAGE_PREFIX = "houston.cloudMigration.";
 export function useMigrationAvailable(): boolean {
   const { data: session } = useSession();
   const gates =
-    isHostedGatewayEngine() &&
-    osIsTauri() &&
-    readMigrationStatus(session) !== "completed";
+    isHostedGatewayEngine() && osIsTauri() && readMigrated(session) !== true;
   const detect = useQuery({
     queryKey: queryKeys.cloudMigrationDetect(),
     queryFn: async () => {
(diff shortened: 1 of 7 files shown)


cravenceiling:
Closing — superseded by #846 (GCP Identity Platform auth migration), which merged after this branch was cut.

#846 retired Supabase and its client-writable `user_metadata`, the exact mechanism this PR's `migrated` flag relied on (`migration-status.ts` was deleted on main). Main already gates the wizard via per-machine localStorage + local `~/.houston` detection: a new user (no old data) goes to onboarding, an existing user (old data present) sees the wizard, and done/skip is remembered per machine — with cross-machine resume handled by the gateway's per-agent import markers.

We're going with main's approach. A server-authoritative `migrated` flag would now have to be a GCP custom claim set backend-side, which we can revisit if that cross-machine guarantee is ever actually needed.

PR #888, opened 2026-07-13
fix(host): fan out runtime transcript changes

cravenceiling:
## Summary

- classify canonical `.houston/runtime/{conversations,sessions}` writes as `ConversationsChanged`
- add an assembled filesystem watcher → `/v1/events` regression test
- update the files-first reactivity documentation

## Root cause

Runtime transcripts moved under `.houston/runtime`, but the shared watcher classifier still matched the deleted legacy paths. Transcript writes were therefore emitted as `FilesChanged` instead of `ConversationsChanged`.

Other connected clients still received `ActivityChanged`, so their composer switched to Stop, but they never invalidated and reloaded the open chat history. The gateway fan-out itself was operating correctly; it was receiving the wrong event vocabulary from the engine pod.

## Verification

- `pnpm --filter @houston/host --filter @houston/runtime --filter @houston/domain test`
- `pnpm --filter @houston/domain --filter @houston/host typecheck`
- `pnpm check`
- `pnpm check:boundaries`
- cloud gateway fan-out/proxy suites: `go test ./internal/edge/events ./internal/proxy`

The new assembled regression test failed before the classifier fix and passes afterward.

### DIFF
--- a/knowledge-base/files-first.md
+++ b/knowledge-base/files-first.md
@@ -16,6 +16,12 @@ Houston uses files, not DB, for agent-visible data. SQLite only for chat replay 
 > read/write; on managed cloud its durable copy lives in the object store, and a
 > fresh pod is expected to boot empty and rehydrate. The layout, atomic-write
 > discipline, schemas, and reactivity model are identical in both worlds.
+>
+> **Canonical chat storage:** v3 conversation transcripts live under
+> `.houston/runtime/conversations/`; pi session state lives under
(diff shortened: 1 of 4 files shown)


cravenceiling:
Native-client follow-up: #889 adds SDK-level `ConversationsChanged` handling so an idle open iOS chat reloads turns completed by another client. #888 merged while that extension was still in progress, so the SDK change is isolated in the follow-up PR.

PR #905, opened 2026-07-14
fix: cloud-migration backup fails on Windows with 'Acceso denegado (os error 5)'

cravenceiling:
## Problem

A Windows user on v0.5.9 hit **`backup copy failed: Acceso denegado. (os error 5)`** on the migration wizard's first step (the local backup of `~/.houston`), which blocks the entire desktop→cloud move.

## Root cause

`copy_dir_all` in `app/src-tauri/src/commands/migration.rs` branched on `DirEntry::file_type()`, which does **not** follow links. A directory symlink or junction inside an agent workspace (pnpm's `node_modules` junction layout is the canonical producer on Windows) therefore fell into the plain-file branch, and `std::fs::copy` on a directory junction fails on Windows with exactly `ERROR_ACCESS_DENIED (os error 5)`, aborting the whole backup.

A second defect made this hard to diagnose: the raw `std::io::Error` carries no filename, so the toast/report showed only the bare os-error string.

## Fix

- `copy_dir_all` now **skips symlinks/junctions** (counted, logged once) instead of trying to copy them. Their targets live outside the entry's own subtree; the real files are still copied from wherever they actually live under `~/.houston`.
- Every IO error in the copy is wrapped with the **offending path**, so any residual failure (e.g. a genuine ACL denial) produces an actionable report: `backup copy failed: C:\Users\...\<file>: Acceso denegado. (os error 5)`.

## Tests

- `copy_dir_all_skips_symlinks` — dir + file symlinks are skipped, real files copied, counts exact.
- `copy_dir_all_errors_carry_the_path` — an unreadable dir yields an error naming the path.
- Existing detection/copy tests updated to the new stats shape. `cargo test --lib`: 145 passed.

### DIFF
--- a/app/src-tauri/src/commands/migration.rs
+++ b/app/src-tauri/src/commands/migration.rs
@@ -151,12 +151,13 @@ pub struct BackupResult {
     pub byte_count: u64,
 }
 
-/// Running totals for [`copy_dir_all`].
+/// Running totals for [`copy_dir_all`]. `skipped` counts entries that are
+/// deliberately not copied: symlinks/junctions and special files.
 #[derive(Default, Debug, PartialEq)]
 struct CopyStats {
     files: usize,
     bytes: u64,
-    skipped_links: usize,
+    skipped: usize,
 }
 
 /// Attach the offending path to an IO error — a bare `std::io::Error` renders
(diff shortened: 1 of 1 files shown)


cravenceiling:
Hardened for macOS too: the identical abort exists there in a different costume — pnpm uses plain symlinks on macOS (`fs::copy` fails with "the source path is neither a regular file nor a symlink to a regular file" instead of os error 5), and a leftover unix socket/FIFO from an agent process would likewise kill the copy. `copy_dir_all` now whitelists regular files + directories and counts/skips everything else, with a test covering dir-symlink, file-symlink, and FIFO.

PR #914, opened 2026-07-15
fix(app): paint sidebar needs-you badges through a pod cold start

cravenceiling:
Follow-up to #911. With the mission cards now painting from the restored caches during a pod cold start, the sidebar's per-agent needs-you badges were still blank until the pods woke (the "Activity 25" tab badge showed, the sidebar pill next to the agent name didn't).

## Root cause

The two badges compute the same number from **different queries**:

- The Activity tab uses the per-agent `["activity", path]` query — a simple, stable key that restores from IndexedDB reliably.
- The sidebar pill derives from the aggregate `["all-conversations", ...agentPaths]` query — and its key embeds **every agent's folderPath**, supplied by the async, **non-persisted** agents roster. TanStack re-attaches a restored entry only on exact serialized-key equality, so while the roster loads (and after any agent add/remove/reorder since the persist) the restored lists sit stranded under a stale key. `NeedsYouChip` renders nothing at count 0, so the sidebar blanks until the live fan-out returns — which a cold pod holds for its whole wake.

## Fix — two placeholder-only layers (the real fetch is untouched)

1. **`useAllConversations`**: when the current key has no data, serve the newest successfully-fetched roster **variant** of the aggregate from the query cache as `placeholderData` (marked `isPlaceholderData`, never persisted, always replaced by the live result). Restored lists now paint regardless of how the roster resolves.
2. **Sidebar summaries**: while the aggregate is not authoritative for the current key, overlay each agent's own cached `["activity", path]` rows via a cache-only `skipToken` subscription (never fetches, so it can't wake a pod), using the same counting rule as the tab badge. The sidebar pill, the board, and the Activity tab then agree by construction during the wake window.

Once the live aggregate lands for the current roster key, it is authoritative again for every agent — behavior after wake is byte-identical to today, including the deliberate never-refetch-on-focus fleet-protection.

## Tests

- `latestCachedAllConversations`: picks the newest successful variant, ignores unrelated keys, empty cache → undefined.
- `summarizeActivities`: same counting rule as the aggregate path (needs_you/running, setup-chat modes excluded).
- Full suite: app 1531 pass, web 187 pass, workspace typecheck + Biome clean.

Verification note: exercised via unit tests + typecheck; the live cold-start window (badges painting while pods wake) needs a hosted build to observe end-to-end.

### DIFF
--- a/app/src/components/shell/agent-activity-summary-model.ts
+++ b/app/src/components/shell/agent-activity-summary-model.ts
@@ -18,6 +18,34 @@ export interface AgentActivitySummary {
   runningCount: number;
 }
 
+/** One agent's board rows (`.houston/activity`), the summary-relevant bits. */
+export interface ActivitySummaryInput {
+  status?: string | null;
+  /** Agent-mode id; routine-setup chats never count toward badges. */
+  agent?: string | null;
+}
+
+/**
+ * Summarize one agent's own activity list — the SAME source (and the same
+ * counting rule) as the "Activity N" tab badge in workspace-shell.tsx, used
+ * as the sidebar fallback while the all-conversations aggregate has not
+ * fetched for the current roster key (cold boot, pods still waking).
+ */
+export function summarizeActivities(
+  activities: ActivitySummaryInput[],
+): AgentActivitySummary {
+  const summary: AgentActivitySummary = { needsYouCount: 0, runningCount: 0 };
+  for (const activity of activities) {
+    if (isSetupChatMode(activity.agent)) continue;
+    if (activity.status === "needs_you") {
+      summary.needsYouCount += 1;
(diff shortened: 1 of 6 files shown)


cravenceiling:
Fixed the e2e failures: the cache-only `useQueries` (skipToken) subscription in the sidebar summaries re-synced its observers on every render, and the notification churn kept the shell re-rendering — the auto-opened setup-chat panel couldn't be Escape-closed (bisected empirically: failures reproduce with the observers attached even when their data is unused, and disappear with them removed). Replaced with `queryClient.getQueryData` reads driven by a raw QueryCache subscription via `useSyncExternalStore` — same badge behavior, no observers, still can't trigger a fetch. Full local Playwright run: 120 passed, 0 flaky.

PR #917, opened 2026-07-15
fix(integrations): Meta Ads cannot connect - never mint a bare custom OAuth auth config

juandavidr7:
## Problem

Connecting **Meta Ads** from the Composio integrations fails when the agent starts (the marketing store agent lists `metaads`): the connect card errors with "can't connect".

## Root cause

Composio has **no managed auth app** for `METAADS` (unlike Gmail/Slack). The unmanaged fallback in `composio-auth-config.ts` took the toolkit's **first** auth scheme, which for Meta Ads is `OAUTH2`, and created a custom auth config with **empty credentials**. A custom OAuth config without a client id/secret can never complete the dance, so connect blew up at agent start. That empty-credentials fallback was only ever valid for user-collectible schemes (API key toolkits like serpapi), where Composio's hosted connect page asks the user for the value.

## Fix

- The fallback now **skips OAuth schemes** (`OAUTH1`/`OAUTH1A`/`OAUTH2`, which need a registered developer app) and picks the first scheme the hosted connect page can collect from the user. For Meta Ads that is `API_KEY`, so connect now works: the hosted page asks the user for their Meta access token.
- An **OAuth-only unmanaged toolkit** now fails loudly naming the remedy: register a developer OAuth app for it in the Composio dashboard; `resolveAuthConfig` reuses that config on the next connect automatically.

## Tests

- New: metaads-shaped toolkit (no managed auth, `[OAUTH2, API_KEY]`) creates a `use_custom_auth` config on `API_KEY`.
- New: OAuth-only unmanaged toolkit rejects with the dashboard remedy in the message.
- `composio.test.ts` 23/23 green; `src/integrations` 104/105 (the one failure asserts POSIX 0600 file modes, impossible on Windows, pre-existing). Biome clean.

## Ops note

If production already persisted an ENABLED `metaads` auth config from the failed attempts (custom OAuth2, empty credentials), `resolveAuthConfig` will keep reusing it - check the Composio dashboard and delete it so the new fallback kicks in. Longer term, registering a Meta developer app in the Composio dashboard gives the proper OAuth UX with zero code changes.


juandavidr7:
This PR now also carries fix(web) 357c712e: the new-engine gate and the Engine Connect screen probed authStatus() against the bare base URL, but the host only serves the pre-agent connect surface under /setup-runtime/* (no flat /auth/status), so connecting to a healthy host failed with 'Can't reach the engine'. Both probes (and the gate's client that ConnectView reuses for the OAuth login routes) now target baseUrl + /setup-runtime; the stored engine config keeps the bare URL.

PR #932, opened 2026-07-16
fix(runtime): expire abandoned OAuth logins so they release the callback port

cravenceiling:
## Problem

An abandoned provider sign-in — started but never completed or cancelled — kept pi's in-process OAuth callback server bound to the provider's **fixed** port forever. For Codex/OpenAI that's `127.0.0.1:1455`, which every other Houston instance and the desktop loopback relay also need. One stale login (e.g. in a forgotten dev runtime) blocked ChatGPT sign-in **machine-wide** until that runtime process died: every later connect failed with *"Couldn't open OpenAI sign-in"* because the port bind collided (`Address already in use`).

Nothing in the runtime bounded an in-flight login's lifetime — only an explicit `cancelLogin` or completion freed the port.

## Fix

`startLogin` now arms a 10-minute abandonment expiry (`LOGIN_TIMEOUT_MS`, matching the local client watcher's cap):

- **On expiry the flow is torn down exactly like `cancelLogin`**: the abort stops device-code pollers, and rejecting the paste promise closes the loopback callback server — freeing the port for the next attempt.
- **The attempt is recorded as an error**, not silently forgotten: the state stays in `active` with the stable `"Login timed out"` sentinel that the frontend already localizes (`PROVIDER_LOGIN_TIMEOUT_ERROR` in `@houston-ai/core`; the runtime is frontend-agnostic, so the string is mirrored by value with a keep-in-sync note). Status polls surface a real toast.
- **Idempotent reuse re-arms the clock**: a second connect click near the deadline gets the full window again, and crossing the original deadline can't kill the refreshed flow.
- The timer is cleared on completion, failure, cancel, logout, and API-key connect; guarded so a stale timer can never clobber a settled or replaced login; and `unref`'d so it never holds the process open.
- After expiry the paste hooks are dropped, so a late paste into a dead dialog throws a loud "no active login" instead of silently resolving into the void.

## Tests

Three new fake-timer tests in `login.test.ts`:
- an abandoned login expires, reports the sentinel, and a retry builds a fresh flow;
- reuse re-arms the abandonment clock (original deadline is defused, refreshed window still expires);
- a cancelled login's stale expiry timer stands down instead of resurrecting an error row.

Full `@houston/runtime` suite: 81 files / 668 tests green. Biome + repo typecheck clean.

### DIFF
--- a/packages/runtime/src/auth/login.test.ts
+++ b/packages/runtime/src/auth/login.test.ts
@@ -2,13 +2,15 @@ import {
   OPENAI_CODEX_BROWSER_LOGIN_METHOD,
   OPENAI_CODEX_DEVICE_CODE_LOGIN_METHOD,
 } from "@earendil-works/pi-ai/oauth";
-import { expect, test } from "vitest";
+import { expect, test, vi } from "vitest";
 import {
   autoPromptAnswer,
   cancelLogin,
   codexLoginMethod,
   getAuthStatus,
   LOCAL_PLACEHOLDER_KEY,
+  LOGIN_TIMEOUT_ERROR,
+  LOGIN_TIMEOUT_MS,
   setApiKey,
   startLogin,
 } from "./login";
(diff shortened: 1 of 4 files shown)


cravenceiling:
Second commit (e5471dcb) rides along: the dev-loop reaper's kill signatures never matched the real host/engine command lines (tsx runs as `cli.mjs`/`loader.mjs`, and the per-agent `packages/runtime` engines weren't listed), so every `pnpm dev` restart leaked the previous session's processes. Those strays are how an abandoned Codex login survived to squat port 1455 machine-wide (the bug the main commit bounds runtime-side), and stale engines also wedged the gateway's ensureAwake hold, hanging the app on the boot splash. The reaper now clears both host forms and the engines.

PR #951, opened 2026-07-16
feat(teams)!: one Permissions view; permission management removed everywhere else

ja-818:
## Element 7 — the permissions end-state (stacked on #948; supersedes the two-lens model from #939/#941/#948)

Felipe's directive, verbatim intent: **one Permissions tab on the menu; inside it People and Agents; nothing else in the app manages permissions.**

### The one place
- **Permissions** (new top-level sidebar view, owner/admin only): tab **People** — each member → which agents they can access (None / Can use / Manager, everyone-shared handled, self-lockout guarded); tab **Agents** — a 'Defaults for every agent' card (the org-wide app + model ceilings) above per-agent cards (integrations ceiling + models ceiling, both default 'everything').

### Zero management anywhere else
- Integrations page = personal connections only. The app detail dialog shows info/reconnect/disconnect — **no agent list, no toggles** (Felipe: 'if I click on an integration it shows all the agents, that's wrong').
- Agent Integrations tab = pure connect surface. The 'Connected, but off for this agent' section is gone. Kept: allowlist transparency + the role-aware **Enable it in Permissions** deep link, and the 'Runs without asking' approvals review (runtime trust, not admin policy).
- Agent Settings loses the apps/models ceiling editors (components live on, mounted only in Permissions).
- Organization dashboard = membership + Activity/Usage + Billing. All policy rows removed.

### The grants layer is dead (this side)
Usable = **connected ∩ effective allowlist**. Deleted end to end: host `LocalIntegrationGrants` + file store + `/v1/agents/:id/integration-grants` routes + sandbox search-filter/execute-403 + custom-integration auto-grant; engine-client/SDK/runtime-client methods; web adapter mixins; app hooks/query keys; fake-host routes. Sandbox now enforces action approvals only. Stale `.houston/integration-grants.json` files are inert. **Sibling cloud PR retires the gateway's C4 grant enforcement** (linked when open).

### Verification
Biome clean (2410 files) · `check:boundaries` OK · host **950** vitest · runtime **674** · fake-host 13 + typecheck · app tsgo + check-locales + **1625** node tests · houston-web typecheck + 188 vitest · e2e **50 green** across permissions-people, permissions-agents (new), integrations-ia (rewritten for the end-state), integrations-locked, ai-models-ia, approval, onboarding-connect. KB: integrations.md §2 rewritten, teams.md rewritten, dispatch table updated.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #952, opened 2026-07-16
feat(teams)!: remove org-wide ceilings; policy is per-agent only

ja-818:
## Element 8 (stacked on #951)

Felipe's call: the global integration + model allowlists were overengineering. Policy is managed **only per agent**.

- Permissions > Agents = the agent list; each agent's card = its integrations + models ceilings (`null` = everything, the default for new agents). The 'Defaults for every agent' card is gone.
- The entire org-settings chain deleted (it held ONLY the two ceilings): `OrgSettings` wire type, engine-client + adapter methods, `useOrgSettings`, editors, fake-host route, i18n blocks.
- Effective allowlist = the agent's own ceiling (no more org ∩ agent). `blockingCeiling` attribution gone — a blocked app is always the agent ceiling, so the 'Enable it in Permissions' CTA always opens that agent's card. The global Integrations page never shows locked rows anymore.
- Per-agent model narrowing (composer clamp, gateway enforcement) unchanged. Historical org-settings audit-log rendering kept (user data).
- Trade-off noted by design: banning an app team-wide = per-agent edits; new agents start fully allowed.

Gateway-side retirement (org ceilings + the already-retired C4 grants) lands in the sibling **cloud** PR (linked when open).

**Verification:** Biome + boundaries clean · app tsgo + check-locales + 1611 node tests · houston-web typecheck · engine-client 71 + fake-host 13 · e2e 16/16 (permissions-people, permissions-agents, integrations-ia, integrations-locked, ai-models-ia). KB + dispatch table updated.


ja-818:
Gateway counterpart is up: gethouston/cloud#132.

PR #978, opened 2026-07-18
fix(local-model): connect directly when the deployment has no tunnel relay

cravenceiling:
## Problem

The guided "connect a local model" flow minted tunnel relay credentials unconditionally, as its very first step. Any deployment without a relay died right there with the generic "Something went wrong while connecting" dialog:

- `pnpm dev`: the dev gateway answers `POST /v1/tunnel/credentials` with 503 "tunnel relay not configured" (confirmed in `~/.dev-houston/logs/frontend.log`)
- a desktop-local engine (no gateway at all): the adapter threw "needs a cloud workspace"
- self-host: same, no relay exists

So local models were only connectable in the hosted cloud, even though in every one of these deployments the engine is co-located with the model server and needs no tunnel at all.

## Fix

- `getTunnelCredentials()` now resolves `null` for every "this deployment has no relay" answer: no control plane, tunnels route absent (404), or relay explicitly unconfigured (the 503). Real failures (auth, transient outage) still throw, so an existing relay that errors surfaces instead of silently downgrading to a direct localhost endpoint against a real cloud pod.
- On `null`, `connectDetectedModel` registers the detected server **directly**: `http://127.0.0.1:<port>/v1`, no bridge, no proxy key (new `buildDirectEndpoint` helper). The engine's save-time validation stays the authority — a managed cloud pod still rejects a localhost URL loudly.
- A saved tunnel bridge whose deployment no longer offers a relay now surfaces a clear reconnect error instead of a raw exception.
- Direct connects save no bridge descriptor, so the tunnel status pill correctly never appears for them (`sessionOwnsBridge` unchanged).

## Tests

- `app/tests/local-model.test.ts`: `buildDirectEndpoint` shape (bare detected origin + `/v1`, no `apiKey`, reasoning/shared toggles carried only when on).
- New `packages/web/tests/tunnel-credentials.test.ts`: the null-classification contract (minted credentials pass through; no-cp, 503-unconfigured, and 404 resolve null; transient 503 and 500 still throw).
- All suites green: Biome, app tsgo + 1662 node tests, web typecheck (shim parity) + 197 vitest tests.

## Verification

In `pnpm dev` with a local server running (LM Studio/Jan/Ollama): open Local model → the detected server connects without the relay, and the chat runs against `127.0.0.1` directly. Hosted cloud is unchanged: relay credentials mint, frpc starts, tunnel pill appears.

### DIFF
--- a/app/src/lib/local-model-connect.ts
+++ b/app/src/lib/local-model-connect.ts
@@ -14,6 +14,7 @@
 import type { CustomEndpoint } from "@houston-ai/engine-client";
 import { showErrorToast } from "./error-toast";
 import {
+  buildDirectEndpoint,
   buildLocalEndpoint,
   type DetectedServer,
   reconnectBridgeArgs,
@@ -111,6 +112,23 @@ export async function connectDetectedModel(opts: {
     const cred = await tauriProvider.getTunnelCredentials();
     if (isAbort(opts.signal)) throw new ConnectAborted();
 
+    // No relay in this deployment (dev, desktop-local engine, self-host): the
+    // engine is co-located with the detected server, so register it DIRECTLY —
+    // no bridge, no proxy key. The engine's save-time validation stays the
+    // authority (a managed cloud pod still rejects a localhost URL, loudly).
+    if (!cred) {
+      await tauriProvider.setCustomEndpoint(
+        buildDirectEndpoint({
+          server: opts.server,
+          model: opts.model,
+          name: opts.name,
+          reasoning: opts.reasoning,
+          shared: opts.shared,
+        }),
+      );
+      return;
+    }
+
     let bridge: Awaited<ReturnType<typeof osStartLocalBridge>>;
     try {
       bridge = await osStartLocalBridge({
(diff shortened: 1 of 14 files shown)


cravenceiling:
Second half of the fix, found by testing the first in `pnpm dev` with `CLOUD_DIR` set: the direct registration then hit the HOST-side managed-cloud egress validation ("Cloud agents can only reach public HTTPS endpoints on port 443", 400) — the dev launcher spawns its engines with `HOUSTON_MANAGED_CLOUD=1`, which also enabled a validation that models a NetworkPolicy dev doesn't have. New `loopbackEgress` dep (set only by `scripts/dev/control-plane.sh` via `HOUSTON_LOOPBACK_EGRESS=1`) skips just that validation while keeping every other managed-cloud behavior; real pods are unchanged and their rejection cases stay pinned by tests.

cravenceiling:
Third commit: the reconnect-card repro (disconnect the local model from AI Models, then send in a chat pinned to it → raw refusal text, no card). The runtime's "No local model configured..." refusal emits no AuthRequired event and wasn't in the feed's auth patterns, so the card never triggered; added the pattern (the card already routes openai-compatible to the local-model dialog). Plus the audit follow-ups: picker treats an "unknown" probe as "checking" not "Not connected", first-load gating uses appears-connected, and a dedicated "make sure the model app is open" card for a mid-turn local endpoint failure (en/es/pt).

Issue #342, opened 2026-05-28
Delay in model picker


cravenceiling:
Sometimes when I start a new chat and open the model picker I only see one AI provider and it appears as "not connected". Then, after a few seconds the other providers appear and models appear as available to choose.

<img width="553" height="306" alt="Image" src="https://github.com/user-attachments/assets/5df92a64-af7e-4e8b-8022-fb5c14911901" />

PR #980, opened 2026-07-18
fix(app): reconnect card for a disconnected local model + unknown-state polish

cravenceiling:
Follow-up to #978 (merged mid-stream, so this ships separately).

## The repro (reported in testing)

Start a chat on the local model, disconnect it in AI Models, send a follow-up: the chat shows the raw runtime refusal ("No local model configured. Set a base URL and model for the OpenAI-compatible provider.") and **no reconnect card**.

Root cause: that refusal emits no `AuthRequired` event (same as the TS engine's "No provider connected" refusal), and the feed's auth-signal patterns didn't include it — so the only trigger for the in-chat reconnect card never fired.

## Fixes

- **Feed pattern** (`provider-auth-feed.ts`): recognize the local-model refusal. The card's existing `openai-compatible` special case then opens the LocalModelDialog (never OAuth), and the confirm probe correctly requires a confirmed `unauthenticated` before showing.
- **Chat model picker** (`model-picker.ts`): an `unknown` probe (engine unreachable / cold pod waking) maps to the neutral `checking` state instead of collapsing to "Not connected" — the #342 flicker, reachable in cloud builds now that the adapter reports `unknown` honestly.
- **First-load gating** (`use-houston-init.ts`): `claudeAvailable` uses appears-connected, so an unknown startup probe no longer degrades gating for a provider that is connected server-side.
- **Mid-turn local endpoint failure** (`provider-error-cards/transient.tsx`): `network_unreachable` for `openai-compatible` gets dedicated copy ("make sure the model app is open and its server is running") instead of the provider-API "check your internet" text — en/es/pt, no status-page button (already suppressed for uncatalogued providers).

Also repairs the stale curated-order expectation in `chat-model-picker-map.test.mjs` (Fable 5 was promoted ahead of Opus 4.8 on main; the `.mjs` test sits outside the `tests/` glob, so it rotted silently — pre-existing failure, confirmed on clean HEAD).

## Tests

New cases: the refusal pattern (verbatim string), `unknown` → `checking` picker mapping (plus confirmed states unchanged). All gates green: app tsgo + 1662 node tests + the out-of-glob `.mjs` picker tests, `check-locales` in sync, Biome clean.

---

## Second round (from live-repro diagnosis)

Testing surfaced that the typed card still didn't appear when the endpoint died **mid-conversation**, plus two more defects found while tracing the session files:

- **Classifier gap (runtime)**: the reconnect-card typing only covered a brand-new conversation (the `getConversation` throw in `chat.ts`). On a CACHED conversation — every message after the first — the same "No local model configured" / "No provider connected" throw comes from `execTurn`'s `resolveModel` and fell through `classifyProviderError` to `unknown`: generic error card, no reconnect flow, no auto-resume. Both guard messages are now `NO_CREDENTIALS_PATTERNS` → `unauthenticated`/`no_credentials`, so the cached path gets the identical typed card and the undelivered-prompt auto-resume.
- **Titles ran on the wrong provider (runtime)**: `summarizeTitle` dispatched on the agent-wide ACTIVE provider, so a local-model chat titled through a full Claude one-shot — a backend the chat never picked, burning a real Anthropic call. Titles now route on the conversation's own provider (`titlePlan`, pure + unit-tested); a disconnected chat provider yields no title (truncation fallback) instead of rerouting. The Claude title one-shot also carries the title instruction in the PROMPT itself: the spawned CLI has been observed running its claude_code preset despite a custom string `systemPrompt`, answering the excerpt so the first line of the answer became the board title ("Yes, I have image capabilities. I…").
- **Silent endpoint wipes now attributed (runtime + host)**: `custom-endpoint.json` was found wiped to `{}` mid-session with no user sign-out. Every wiper funnels through the runtime's `logout("openai-compatible")` or the host's org-share clear, and both were silent. Each now logs loudly, so the next wipe names its caller in the runtime log.

Tests: classifier cases for both guard messages, `titlePlan` routing (local chat titles on the local model, anthropic chat keeps its own model, uncached falls back to active), title-instruction-in-prompt. Runtime 693 + host 971 green, Biome clean, workspace typecheck clean.

### DIFF
--- a/app/src/components/shell/provider-error-cards/auth.tsx
+++ b/app/src/components/shell/provider-error-cards/auth.tsx
@@ -31,10 +31,12 @@ import { CheckCircle2Icon } from "lucide-react";
 import { useEffect, useRef, useState } from "react";
 import { useTranslation } from "react-i18next";
 import { subscribeHoustonEvents } from "../../../lib/events";
+import { getProvider } from "../../../lib/providers";
 import { tauriProvider } from "../../../lib/tauri";
 import { useUIStore } from "../../../stores/ui";
 import { RowCard } from "../../cards/row-card";
 import { RowCardButton } from "../../cards/row-card-button";
+import { LocalModelDialog } from "../local-model-dialog";
 import { ProviderGlyph } from "../provider-logos";
 import {
   type AuthCardButton,
@@ -57,6 +59,7 @@ export function UnauthenticatedCard({
   const [launching, setLaunching] = useState(false);
   const [failureDetail, setFailureDetail] = useState<string | null>(null);
   const [retrying, setRetrying] = useState(false);
+  const [showCustomDialog, setShowCustomDialog] = useState(false);
   const relaunchingRef = useRef(false);
   // The auto-resend must fire ONCE per card — a second fire (the provider can
   // complete several logins while the chat stays open) would double-send.
@@ -102,6 +105,17 @@ export function UnauthenticatedCard({
 
   const reconnect = async () => {
     if (launching) return;
+    // A local OpenAI-compatible server has no OAuth flow — reconnect by
+    // re-connecting the local model in its dialog, not launchLogin (which
+    // would throw "does not use OAuth sign-in" and dead-end the card). A
+    // successful connect fires the same `ProviderLoginComplete` event
+    // (`setProviderCustomEndpoint`), so the auto-resume above still runs.
+    if (error.provider === "openai-compatible") {
+      setFailureDetail(null);
+      setShowCustomDialog(true);
+      setPhase("waiting");
+      return;
+    }
     setLaunching(true);
     relaunchingRef.current = true;
     setFailureDetail(null);
@@ -118,6 +132,12 @@ export function UnauthenticatedCard({
   };
 
   const cancelSignIn = async () => {
+    // Local model: nothing engine-side to cancel — close the dialog and re-arm.
+    if (error.provider === "openai-compatible") {
+      setShowCustomDialog(false);
+      setPhase("idle");
+      return;
+    }
     try {
       await tauriProvider.cancelLogin(error.provider);
     } finally {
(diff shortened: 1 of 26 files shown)


cravenceiling:
Fourth commit — the real root cause of the stuck chat, traced on a live repro: the send WAS reaching the pod; the runtime's pre-execution failure path (pinned local model, endpoint disconnected) persisted the typed providerError and published the stamped `error` frame but never the nonce-stamped user echo — and per the turn-identity contract the sink drops a stamped frame when it has no adopted id. The client dropped its own terminal frame: spinner forever, no feed error, no reconnect card. Fixed in the runtime (failure path now publishes echo-then-error like a normal turn) and hardened in the SDK (an accepted send adopts a stamped terminal frame when no id was adopted — also fixes older pods). Feed-pattern fix from the earlier commit then triggers the card off the settled error.

cravenceiling:
Fifth commit — the actual root cause of the indefinite spinner, isolated by tracing a live repro end-to-end and reproducing it as a failing unit test.

The earlier commits made a *live* subscriber settle. But the failure is instant: the pod publishes the terminal error and **clears its replay buffer** before the turn's own subscription attaches. I proved the pod cannot re-serve those frames — `?after=<seq>` past a terminal returns a bare `resync` sync, never the lost error frame. The client's backstop (the pre-settled poll) only armed after seeing a fresh idle *sync* — so when a flaky SSE hop delivers **no frame at all**, the poll never armed and the turn spun until the ~6-minute reconnect budget expired. A unit test feeding `sendAccepted()` + zero frames confirmed `settled: false`.

Fix: arm the conclusive-only poll on an accepted send alone. It settles only on a real persisted reply, so a running turn re-arms harmlessly and a running sync cancels it — normal turns pay nothing. The failed turn now settles from its persisted `providerError`, which the feed pattern surfaces as the reconnect card.

cravenceiling:
Sixth commit — the wrong-provider reconnect card ("Sign in with Anthropic" on a Jan chat) + the auto-resume ask. The store card derives its provider from the chat resolution chain, and the activity record carried stale agent defaults (anthropic/claude-sonnet-5) while every turn ran on Jan. Fix: the runtime now types the not-connected failure `unauthenticated` with `undelivered_prompt` (HOU-718's existing contract), so the INLINE typed card renders with the provider the turn actually failed on; its button opens the guided LocalModelDialog for `openai-compatible` (ProviderLoginComplete then auto-resumes the task by re-delivering the undelivered prompt); and the store card is suppressed whenever an inline auth card exists, so a stale resolution can never render a second wrong-provider card.

PR #985, opened 2026-07-19
feat(app): send app-version header to the gateway + blocking required-update screen

cravenceiling:
Client half of the hosted gateway's minimum-app-version floor (PR 3 of the update-management rollout; the server side is already live and dark — no floor configured).

## What

- **Identity header.** Every gateway request carries `X-Houston-App-Version: <semver>+<channel>`, injected at `gatewayAuthFetch` — the single choke point all gateway HTTP rides (authFetch, cpFetch, SDK, per-agent + setup-runtime clients, boot probes). Channel (`cloud`/`local`) derives from the **same** baked flag that makes a build a cloud build (`VITE_HOSTED_ENGINE_URL` via `resolveEngine`) — the exact builds whose updater `point-updater-at-cloud-manifest.sh` repoints — so channel and updater feed cannot drift. No build-script or Rust changes.
- **Web stays out of scope structurally.** The shared adapter reads window globals (`__HOUSTON_APP_VERSION__`, `__HOUSTON_UPDATE_REQUIRED__` — the `__HOUSTON_SESSION_REFRESH__` idiom) that only the desktop shell installs at module load; the web bundle installs neither, so it never sends the header and can never trip the blocking screen. The local sidecar ignores the unknown header.
- **Required-update state.** Two triggers feed a latched pub/sub bus: any gateway **426** (body's `minVersion`/`updateUrl` parsed off a clone; the response is returned unchanged so callers keep their error path), and a `minAppVersion` on gateway `/v1/version` above this build (one probe per run at connect; the route is 426-exempt so a locked-out app can still read its floor). Signals merge — a later signal can't erase a known `minVersion`/`updateUrl` — and never clear: only installing + relaunching satisfies the floor. The bus latches so a 426 during boot is replayed to the late-mounting shell.
- **Blocking screen.** `UpdateRequired` replaces the dismissible card when required: full-window, non-dismissible (`role="alertdialog"`), shows current → minimum version, drives the existing updater flow (install → progress → relaunch → error/retry), falls back to opening the gateway-provided `updateUrl` when the updater can't serve (dev/feed unreachable), and to a re-check button when even that is absent — never a dead end. Shown on hosted-gateway builds only: a local/sidecar build never has the gateway on its request path, so local-only work can't be soft-locked.
- Version compare is numeric-triple-only, fail-open on anything unparseable (a `0.5.9-dev` build whose numbers meet the floor must not self-block; the gateway's 426 stays the authority). Added an `update_required` analytics event alongside the existing update funnel.

## Tests

19 new tests: channel/header derivation per build flavor, triple-compare semantics (incl. prerelease + fail-open), bus latching/merging, `gatewayAuthFetch` header injection + 426 forwarding on both auth attempts with the body left unconsumed, `minAppVersion` extraction. Full app suite 1676/1676, web suite 207/207, repo-wide typecheck, biome, and en/es/pt locale parity all green.

### DIFF
--- a/app/src/components/shell/update-checker.tsx
+++ b/app/src/components/shell/update-checker.tsx
@@ -3,13 +3,36 @@ import { useTranslation } from "react-i18next";
 import houstonBlack from "../../assets/houston-black.svg";
 import houstonWhite from "../../assets/houston-icon-white.svg";
 import { useUpdateChecker } from "../../hooks/use-update-checker";
+import { isHostedGatewayEngine } from "../../lib/engine";
 import { selectUpdateNotes } from "../../lib/update-details";
 import { UpdateNotes } from "./update-notes";
+import { UpdateRequired } from "./update-required";
 
 export function UpdateChecker() {
   const { t, i18n } = useTranslation("shell");
-  const { status, installAndRelaunch, relaunchInstalledApp, dismiss } =
-    useUpdateChecker();
+  const {
+    status,
+    required,
+    installAndRelaunch,
+    relaunchInstalledApp,
+    dismiss,
+  } = useUpdateChecker();
+
+  // Hard floor: the hosted gateway refused this build, so gateway calls are
+  // failing 426 — replace the dismissible card with the blocking screen.
+  // Scoped to hosted-gateway builds: on a local/sidecar build every feature is
+  // served by the co-located host (the gateway is never on the request path),
+  // so even a stray floor signal must not lock the user out of local work.
+  if (required && isHostedGatewayEngine()) {
+    return (
+      <UpdateRequired
+        required={required}
+        status={status}
+        onInstall={installAndRelaunch}
+        onRelaunch={relaunchInstalledApp}
+      />
+    );
+  }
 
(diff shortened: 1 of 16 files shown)


cravenceiling:
**Fix pushed for the Web e2e failure** (`e22dd29`): the shared shell's `engine.ts` is bundled into the web app too, so the module-load bridge install leaked the `X-Houston-App-Version` header into browser builds — and a custom header makes every fetch CORS-preflighted, which the engine host's / fake host's static `Access-Control-Allow-Headers` rejected: every request failed and the shell never booted (deterministic global-setup timeout).

- Bridge install is now gated on `osIsTauri()` — a browser tab never identifies as an app build (which also restores the "web out of scope" invariant by mechanism, not luck).
- Engine host + fake host now allow the header for the desktop webview's preflights against the sidecar.
- ⚠️ **Merge-order dependency:** the hosted gateway's own CORS allowlist needs `x-houston-app-version` (companion cloud PR) **deployed before any app release ships this header** — a disallowed header fails the preflight and takes every desktop→gateway request down with it, floor configured or not.

PR #999, opened 2026-07-20
fix(onboarding): migration-flow polish, durable onboarding-completed signal, deleted-account sign-out

ja-818:
## What

Three fixes around the cloud-migration / first-run flow (HOU-719/732 family):

1. **Migration wizard UI** — the Connect-your-AI step now renders the same curated provider view as regular onboarding (explicit Continue, no auto-advance); Reconnect-your-apps is a 2-column grid and connected apps show a success check instead of a dimmed disabled button.
2. **Durable `onboarding_completed` engine pref** — zero agents no longer conflates a fresh install with an emptied workspace or a finished migration. Set on both onboarding terminal paths, on migration outcome "done", backfilled for existing users; uid-keyed localStorage mirror so a pod-pref blip can never re-onboard a real user. Routing extracted to pure `onboardingRoute()` with table-driven tests. Result: migration done → straight to the shell; delete-all-agents → the shell's empty state; fresh install unchanged (segmentation included).
3. **Deleted-account sign-out** — a `USER_NOT_FOUND` token refresh (account deleted server-side) is now a terminal sign-out instead of an infinite retry that traps the app behind 401s.

## Testing

- `onboarding-flow` + `identity-refresh` node test suites (30 passing), app + web typecheck, Biome clean.
- Reproduced end-to-end against the local dev stack with a from-zero test account.

Sibling PR (cloud): the gateway-side setup-agent list fix + the scoped test-user reset tool — see gethouston/cloud PR referenced below.


ja-818:
Sibling cloud PR: https://github.com/gethouston/cloud/pull/150

PR #1001, opened 2026-07-20
feat(app): legacy→cloud migration offer for the final local release

cravenceiling:
Implements the migration plan for legacy (local-engine) users: this is intended to ship as the **final local-channel release** — its whole feature is the offer to upgrade to Houston Cloud.

## What users see (local/sidecar builds only; cloud builds render nothing)

- **Announcement modal on every launch** (per product decision): hero header, centered copy, single pill CTA — modeled on the reference screenshot. Copy states it's an upgrade, that Houston Cloud is **early beta**, that agents/data **import automatically**, and that integrations must be **reconnected** after signing in.
- **Not pushy**: the X only hides it for the session, and the copy points at the always-available reopen paths — a **sidebar entry "Migrate to cloud"** and a **settings row** — both of which just reopen the same modal.
- Install shows progress → "Relaunch" boots the cloud build → the existing HOU-719 first-run wizard handles data import + integration reconnect. Nothing new to design post-install.

## How the cross-channel install works

- New Rust command `install_cloud_migration`: `UpdaterBuilder` pointed at the **cloud** manifest (`cloud-latest/latest-cloud.json`) with `version_comparator` forced true (channels are versioned independently). JS `check()` can't override endpoints, hence Rust. Signature verification is inherited — both channels sign with the same key. Progress rides a `cloud-migration-progress` event (same pattern as the dictation model download).
- Relaunch reuses the existing `current_app_bundle_path` / `relaunch_app_from_path` choreography.

## The future "force" lever (built now, because there is no later)

`fetch_migration_policy` (Rust, CORS-free) reads `releases/download/migration-policy/migration-policy.json`. `{"mode":"required"}` removes the X and the offer becomes a hard gate; anything else — including 404/network failure — is **optional** (fail-open). Flipping it later is one `gh release upload --clobber`, no build.

## Also

- `migrate_to_cloud_{offered,accepted,dismissed,installed}` analytics events (adoption funnel measurable in PostHog).
- Web-build shims for both new commands (parity check enforced by pre-commit).
- en/es/pt strings; locale parity check passes; typecheck, biome, and `cargo check` green.

## Release sequencing (after merge — separate step, needs sign-off)

Cut as `v0.4.29` (**below** the cloud line's 0.5.19): local users on 0.4.28 auto-update into it via the normal published-release flow, and the Windows MSI upgrade guard can never block the subsequent legacy→cloud install (cloud version stays higher). Publishing a `v*` release also promotes the web bundle to production (`web-promote`) — flagged so it's a conscious side effect.

### DIFF
--- a/app/src-tauri/src/commands/update.rs
+++ b/app/src-tauri/src/commands/update.rs
@@ -6,6 +6,101 @@
 //! asks this module to open that path after install.
 
 use std::path::{Path, PathBuf};
+use tauri::{AppHandle, Emitter};
+use tauri_plugin_updater::UpdaterExt;
+
+/// The CLOUD channel's rolling updater manifest (kept current by
+/// cloud-updater-manifest.yml on every published cloud-v* release). The
+/// local build's baked endpoint is the LOCAL feed (latest.json); migration
+/// deliberately crosses channels, so the endpoint is overridden here — the
+/// JS `check()` cannot do that (no `endpoints` in CheckOptions).
+const CLOUD_MANIFEST_URL: &str =
+    "https://github.com/gethouston/houston/releases/download/cloud-latest/latest-cloud.json";
+
+/// Remote migration policy: `{"mode":"optional"}` or `{"mode":"required"}`.
+/// Lives as an asset on the fixed `migration-policy` release so the mode can
+/// be flipped later with one `gh release upload --clobber` — this is the LAST
+/// local-channel build, so any future behavior change must ride a remote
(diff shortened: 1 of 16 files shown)


cravenceiling:
Rebasing this onto the legacy release line per Daniel: the legacy app ships from the release-0.4.26 branch (v0.4.28 was cut there), not main — and its final release should carry ONLY the migration offer, not main's 492-commit delta. Replacement PR targeting release-0.4.26 incoming.

PR #1003, opened 2026-07-20
feat(app): legacy→cloud migration offer (final 0.4.x feature)

cravenceiling:
The migration plan for legacy users, built **on the legacy release line** (`release-0.4.26`, whose tip is the v0.4.28 build users actually run) — main is not involved. This is intended to ship as **v0.4.29, the final release of this line**: its only change is the offer to move to Houston Cloud.

## What users see

- **Announcement modal on every launch** (astronaut split layout: full-height photo left, benefit rows + amber reconnect warning right, single pill CTA): free-for-early-believers framing, agents/data **import automatically**, integrations must be **reconnected** after signing in.
- **Explicitly not pushy**: the copy says there's no need to decide now, and points at the always-available **sidebar entry "Move to the cloud" (rocket icon)** (plus a settings section) — both just reopen the same modal. The X hides it for the session only.
- Dev builds skip the auto-open (`import.meta.env.PROD` gate) so `pnpm dev` isn't greeted by the modal; sidebar/settings still open it for testing.
- Install shows progress → Relaunch boots the **cloud** build → its first-run migration wizard handles the ~/.houston import + reconnect flow.

## Mechanics

- New Rust command `install_cloud_migration`: updater pointed at `cloud-latest/latest-cloud.json` with `version_comparator` forced true (channels are versioned independently; this line sits at 0.4.x below cloud's 0.5.x, which also keeps the Windows MSI upgrade guard permissive). Signature verification inherited — both channels sign with the same key. Progress rides a `cloud-migration-progress` event.
- `fetch_migration_policy` (Rust, CORS-free) reads `releases/download/migration-policy/migration-policy.json`: `{"mode":"required"}` removes the X (hard gate); anything else including 404/network failure is **optional** (fail-open). Flipping it later is one `gh release upload --clobber` — no further 0.4.x build needed, which matters because there won't be one.
- `migrate_to_cloud_{offered,accepted,dismissed,installed}` analytics; en/es/pt strings (locale parity + no-em-dash check pass).

## Verification

- `tsc --noEmit` clean, `cargo check` clean (with the `houston-engine` sidecar staged), app tests 355/355, `check-locales` green — all against this 0.4.x toolchain.

## Release sequencing (after merge, separate sign-off)

Tag `v0.4.29` from this branch (bump version files on the release commit). Publishing it delivers this build to all 0.4.x users through their normal auto-update, and the modal starts greeting them. Note: publishing a `v*` release also re-runs `web-promote` from this branch's web bundle — same-vintage as what production already serves, so a no-op in practice.


cravenceiling:
Added in 73096fef: **OS-trust-store TLS fix** (Sentry HOUSTON 7541949805 — `UnknownIssuer`, 377 events since June, user stuck on 0.4.26). The bundled webpki roots reject TLS-intercepting AV/corporate proxies whose CA sits in the OS store; that silently breaks the 30-min update check (users stop receiving updates) and would equally break this PR's cloud-migration download. `tauri-plugin-updater` gains `native-tls` (schannel on Windows / Secure Transport on macOS; defaults incl. `zip` kept — with both stacks compiled, reqwest prefers native-tls), and the app's own reqwest moves to `rustls-tls-native-roots`. Known limitation: users whose updater is already hard-broken by this won't receive 0.4.29 automatically — the fix rides the channel that's broken for them — but intermittent cases recover, and everyone who gets 0.4.29 gets a working migration download.

PR #1045, opened 2026-07-23
feat(chat): upload whole folders as chat attachments (HOU-808)

cravenceiling:
Fixes HOU-808 — a user asked to upload folders of files in chat; until now the composer only took individual files, and dropping a folder silently attached nothing.

## Root cause

Three layers each blocked folders:
- the composer's hidden `<input type="file">` had no directory option;
- both drop zones read only `dataTransfer.files`, which flattens a dropped folder to an empty list — the drop did nothing, with no feedback;
- the host's `POST /agents/:id/attachments` rejected any filename containing a path separator, so folder structure had nowhere to go.

## What changed

**Host** (`packages/host/src/turn/attachments.ts`): each uploaded file may carry a `relPath` (mirrors `File.webkitRelativePath`, e.g. `docs/guide/a.md`). Every segment is held to the existing `safeFilename` rules (no traversal, no separators, no dot-names), capped at 32 segments / 1024 chars, and the file lands at `uploads/<relPath>` with the same per-file " (n)" dedupe as before. Folders deliberately **merge** rather than dedupe at the folder level: the client uploads one folder across several batched requests, and the returned per-file paths are exact either way.

**Adapter** (`packages/web/src/engine-adapter/cp/files-context.ts`): forwards `webkitRelativePath` as `relPath` and batches small files (8 MB / 25 files per request) instead of one request per file, so a 200-file folder isn't 200 round trips. Engine pods predating the field ignore it and store flat filenames — graceful degradation, no version gate.

**ui/chat**:
- "Add a folder" entry in the composer attach menu, backed by a second hidden `webkitdirectory` input;
- folder drag-and-drop via `webkitGetAsEntry()` traversal (entries captured synchronously before the `DataTransfer` is neutered; `readEntries` drained in batches); files produced by traversal are tagged with the same `webkitRelativePath` shape the picker produces;
- an attached folder renders as **one chip** (folder icon, "N files", one-click remove of the whole folder);
- hidden files (`.DS_Store`, `.git/**`) are filtered out of folder picks and drops — the host refuses dot-named segments, and failing the whole send over an invisible file would be worse;
- a 1000-file ceiling fails loudly with a notice (dropping `node_modules` must not upload for an hour);
- dedupe identity now includes the relative path, so two same-named files in different subfolders both attach.

**Prompt** (`app/src/lib/attachment-message.ts`): a folder collapses to one line — `- uploads/docs/ (uploaded folder with 12 files inside)` — instead of hundreds of file lines; the agent lists the folder with its file tools on demand. Only relative `uploads/…` paths group; legacy absolute paths keep per-file lines.

i18n: new strings in en/es/pt (`composerAttach.addFolder`, `composer.tooManyFiles`, `composer.folderReadFailed`, `composer.folderFileCount`).

## Reproduce the bug (before)

1. `pnpm dev`, open an agent chat.
2. Click the + attach button → only "Add files"; the picker cannot select a folder.
3. Drag a folder from Finder onto the chat: the overlay shows, the drop does nothing — no chip, no error.

## Verify the fix (after)

1. `pnpm dev`, open an agent chat.
2. Click + → **Add a folder**, pick a folder with subfolders: one folder chip appears with the file count.
3. Send with a prompt like "what's in there?": the message shows "N files attached"; the agent reads `uploads/<folder>/…` with structure intact; the Files tab shows the nested folder.
4. Drag-and-drop a folder onto the chat: same result.
5. Drop a mixed selection (folder + loose files): loose files chip individually, the folder stays one chip.
6. Attach the same folder twice: files inside dedupe as `name (1).ext`, still inside the same folder.
7. Drop a huge folder (>1000 files): loud notice, nothing attached.

Tests: host attachments suite grown to 18 (structure, merge/dedupe, hostile relPaths, HTTP relPath, non-string relPath 400), app prompt-grouping tests. Full gates green: host 1048 / runtime 737 / domain 147 / app 1957 tests, `pnpm typecheck`, `packages/web` typecheck + shim parity, `check-locales`, `check:boundaries`, Biome.

### DIFF
--- a/app/src/components/board/mission-control-archived.tsx
+++ b/app/src/components/board/mission-control-archived.tsx
@@ -142,6 +142,7 @@ export function MissionControlArchived({
           }
           onPanelOpenChange={setMissionPanelOpen}
           onOpenLink={openHref}
+          onNotice={(message) => addToast({ title: message })}
           prepareAttachments={attachmentValidation.prepareAttachments}
           onAttachmentRejections={attachmentValidation.onAttachmentRejections}
           thinkingIndicator={panel.thinkingIndicator}
--- a/app/src/components/board/use-board-labels.ts
+++ b/app/src/components/board/use-board-labels.ts
@@ -1,4 +1,5 @@
 import type { AIBoardProps } from "@houston-ai/board";
+import { MAX_ATTACHMENT_FILES } from "@houston-ai/chat";
 import { useTranslation } from "react-i18next";
 
 /**
@@ -30,6 +31,12 @@ export function useBoardLabels(): {
       dropTitle: t("chat:composer.dropTitle"),
       dropDescription: t("chat:composer.dropDescription"),
       imagePasteUnavailable: t("chat:composer.imagePasteUnavailable"),
+      tooManyFiles: t("chat:composer.tooManyFiles", {
+        max: MAX_ATTACHMENT_FILES,
+      }),
+      folderReadFailed: t("chat:composer.folderReadFailed"),
+      folderFileCount: (count: number) =>
+        t("chat:composer.folderFileCount", { count }),
     },
   };
 }
--- a/app/src/components/tabs/archived-tab.tsx
+++ b/app/src/components/tabs/archived-tab.tsx
@@ -34,6 +34,7 @@ export default function ArchivedTab({ agent, agentDef }: TabProps) {
   const panelContainer = useDetailPanelContainer();
   const { data: rawItems } = useActivity(path);
   const deleteActivity = useDeleteActivity(path);
+  const addToast = useUIStore((s) => s.addToast);
   const setMissionPanelOpen = useUIStore((s) => s.setMissionPanelOpen);
   const viewMode = useUIStore((s) => s.viewMode);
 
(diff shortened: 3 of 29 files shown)


cravenceiling:
Follow-up commit addresses an independent review pass:

- **No partial persist**: the host now validates the whole attachment batch before writing anything — one bad path 400s the request with zero files stored (new test).
- **No stale drop closure**: async folder traversal delivers through a latest-callback ref, so a slow traversal can't overwrite newer attachment state or resurrect cleared attachments.
- **Fail fast on huge dirs**: the directory drain aborts at the 1000-file ceiling instead of enumerating everything first.
- **No silent drop failures**: archived tab, mission-control archived and routine intake composers now toast notices; the chat panel rethrows (→ Sentry) when no notice channel exists.
- **Symlink note**: lexical relPath validation can't stop the fs from following a symlinked dir under `uploads/`. Deliberately accepted and documented in code: only a principal that already writes to the workspace (the agent itself) can plant one, and it can already write wherever the link points — same posture as the pre-existing flat uploads, and the Vfs abstraction (GCS included) has no lstat.

PR #1051, opened 2026-07-23
perf(chat): open chats on the transcript tail + scroll-up lazy-load (HOU-819)

cravenceiling:
Fixes HOU-819 — "Chats hang for a while before opening or accepting input."

## Root cause

Opening a chat fetched, folded, and rendered the **entire conversation transcript**, every time:

- `GET /conversations/:id/messages` had no pagination anywhere in protocol v3 — the runtime reads and JSON-parses the whole per-conversation file and ships every message.
- The frontend folded and rendered all of it at once (no windowing, no virtualization), blocking the main thread — the composer isn't disabled during this, but the UI can't respond, which reads as "won't accept input."
- Worst amplifier: every `ConversationsChanged` event invalidates the open chat's history query, so **during a live turn the full-history fetch + fold re-fired repeatedly**.
- On cloud, that unpaginated read is also the request the gateway holds for the whole engine-pod cold start.

## Fix — window the transcript

- **protocol**: `ConversationHistory` gains additive `offset` + `totalMessages`.
- **runtime**: `GET /conversations/:id/messages?limit=N&before=M` serves the tail (or the page ending at absolute index `M`). No params → full history, byte-compatible for old clients; a proxy that strips query strings degrades to today's behavior, never breaks.
- **sdk**: the conversation VM gains additive `historyWindow` state plus `prependHistory` / `stampHistoryWindow` seams.
- **web adapter**: chat opens fetch the **last 120 messages** (bulk/mission-search reads keep the full fetch); a recency-based seed guard (`decideServerSeed`) replaces the old length guard so a shorter-but-newer server tail still lands over a stale cache paint; the local cache now carries frame timestamps and caps at the open window; `loadOlderChatHistory` prepends 80-message pages (single-flight, drops a page that raced a reseed).
- **ui/chat + ui/board + app**: a `ConversationLoadOlder` sentinel at the top of the feed auto-loads older pages as the user scrolls up, viewport-anchored so on-screen content never jumps. Wired on the per-agent mission board, Mission Control, and both archived surfaces.
- **inventory v34**: `conversation-feed` gains `load-older-trigger` / `loading-older`.

The invalidation storm is defanged as a side effect: each refetch is now a bounded tail read, and identical-content revalidations stamp the window without reseeding (no feed id churn / remount).

## Reproduce (before)

1. Use an agent with a long mission (several hundred messages — long tool-heavy runs get there fast).
2. Click its mission card: the panel opens but the transcript takes seconds to paint, and the window is unresponsive to typing while the full transcript folds/renders.
3. Send a message and watch the network: every `ConversationsChanged` during the turn re-fires the full `GET …/messages` for the open chat.
4. On cloud with a cold pod, the first open additionally hangs on the held full-history read.

## Verify (after)

1. Same long-mission card: the chat opens on the last 120 messages near-instantly and typing works immediately.
2. Scroll to the top of the feed: a quiet spinner appears and the previous 80 messages prepend with no viewport jump; repeat until the transcript start, where the trigger disappears.
3. During a live turn, network shows only bounded `?limit=120` reads (and only when the fold actually changed does the feed reseed).
4. Old client ↔ new server and new client ↔ old server both still work (params ignored / fields absent → full-history behavior).

Gates: `pnpm check` clean, `pnpm typecheck` (ui/sdk/app/web + shim parity), `pnpm check:boundaries`, `pnpm check:parity` (inventory v34), vitest: runtime 743, sdk 375, web 218 — all passing.

### DIFF
--- a/app/src/components/board/board-source.ts
+++ b/app/src/components/board/board-source.ts
@@ -119,6 +119,11 @@ export interface BoardSource {
     sessionKey: string,
     opts?: HistoryLoadOptions,
   ) => Promise<FeedItem[]>;
+  /** Scroll-up lazy-load for the OPEN chat (HOU-819): prepend the previous
+   *  transcript page of the active conversation. */
+  onLoadOlderMessages?: () => Promise<unknown>;
+  /** Older messages exist beyond the open chat's loaded window. */
+  hasOlderMessages?: boolean;
   /** Raw send (no queue). `overrides` carry the composer's effective
    *  provider/model; the per-agent source uses them, Mission Control resolves
    *  its own from the target activity. */
--- a/app/src/components/board/mission-board.tsx
+++ b/app/src/components/board/mission-board.tsx
@@ -146,6 +146,8 @@ export function MissionBoard({ source }: { source: BoardSource }) {
           onRemoveQueuedMessage={sendQueue.onRemoveQueuedMessage}
           queuedLabels={queuedLabels}
           onLoadHistory={source.loadHistory}
+          onLoadOlderMessages={source.onLoadOlderMessages}
+          hasOlderMessages={source.hasOlderMessages}
           onNewPanelOpenerReady={source.registerOpener}
           onPanelCloserReady={handleCloserReady}
           emptyState={source.emptyState}
--- a/app/src/components/board/mission-control-archived.tsx
+++ b/app/src/components/board/mission-control-archived.tsx
@@ -134,6 +134,8 @@ export function MissionControlArchived({
           onSendMessage={handleSendMessage}
           onComposerSubmit={panel.onComposerSubmit}
           onLoadHistory={data.loadHistory}
+          onLoadOlderMessages={data.onLoadOlderMessages}
+          hasOlderMessages={data.hasOlderMessages}
           emptyState={
             <ArchivedEmptyState
               hasQuery={missionSearch.hasQuery}
--- a/app/src/components/board/use-agent-board-data.ts
+++ b/app/src/components/board/use-agent-board-data.ts
@@ -9,7 +9,7 @@ import {
   useDeleteActivity,
   useUpdateActivity,
 } from "../../hooks/queries";
-import { useConversationFeed } from "../../hooks/use-conversation-vm";
+import { useConversationVm } from "../../hooks/use-conversation-vm";
 import { useWarmingBoardRows } from "../../hooks/use-warming-board-rows";
 import { missionCardTags } from "../../lib/mission-card";
 import { canDropMission, selectActive } from "../../lib/mission-selection";
@@ -22,6 +22,8 @@ import type { Agent, AgentDefinition } from "../../lib/types";
 import { mergeWarmingRows } from "../../lib/warming-board-rows";
 import { useUIStore } from "../../stores/ui";
 import { missionColumnIdForStatus } from "../mission-board-columns";
+
+const EMPTY_FEED: FeedItem[] = [];
 
 /**
  * Per-agent board data: maps this agent's activities to kanban items, exposes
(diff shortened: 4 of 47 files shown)


cravenceiling:
Adversarial review pass (Codex, high effort) found three real issues in the first commit; all addressed in c372eb9e:

1. **Cross-clock timestamp comparison** in the windowed-seed guard (client-stamped cached frames vs runtime-stamped history) could suppress the load-older trigger after any live session. Replaced with content-anchored decisions on the user-message turn skeleton — superset detection, identical-content stamping, loaded-pages preservation, and an optimistic-send guard (11 decision-table tests).
2. **Cache cap counted frames, not messages**, so a tool-heavy turn could be cut apart and cold-open a transcript missing its own prompt. The trim now walks back to the previous turn boundary (360-frame budget, 600 hard ceiling).
3. **Scroll anchoring overshoots** if an actively streaming turn grows the feed during a load-older fetch — documented as a known limitation; the proper fix needs stable feed-entry React keys (message keys are position-based today, which also means each prepend re-keys the list below). Follow-up candidate, along with virtualizing the message list and making the runtime's history/list reads async (they're synchronous whole-file reads on the event loop).

cravenceiling:
Verified live end-to-end (dev host on :4319 + web UI + Playwright) against a real dev conversation seeded to 312 messages:

- **Server**: full fetch stays byte-compatible (312 msgs, offset 0); `?limit=120` returns exactly the last 120 (offset 192); pages chain gap-free 192→112→32→0; `before` clamps at both ends; malformed params fall back to full history.
- **UI**: chat opens on 120 rendered messages with the window boundary exact; wheel-scrolling to the top auto-fires `?limit=80&before=…` pages until the transcript start, where the trigger retires.
- The live run caught one more real bug: message React keys were **position-based**, so a prepend re-keyed the list and reused Streamdown instances against the wrong messages — the middle of the transcript rendered stale (the #364 failure class, triggered by prepend instead of session switch). Fixed in 825d95ed: `FeedItem` carries the VM feed-entry id and messages key off it (positional fallback for id-less consumers). Re-verified: all 312 messages render correct content after a full scroll-up drain.

Stable keys also mean prepends no longer remount the messages below, and they unblock the anchor-node scroll fix for the streaming-concurrent case noted above.

cravenceiling:
Follow-up performance layers added in e23a7de5 (all verified live against a 312-message conversation through the real host + UI):

1. **Message-list rendering virtualization** — `content-visibility: auto` + `contain-intrinsic-size` on every message root. The browser lays out/paints only the viewport slice: measured **4 of 120** messages rendered on chat open, **52 of 312** after fully draining the transcript via scroll-up. Offscreen messages render on demand (scroll-into-view verified), scroll-to-bottom and load-older anchoring unaffected, and engines without support ignore the properties entirely.
2. **Runtime store parse cache** — an mtime/size-validated LRU behind `loadConversation` and the list scan. History reads hit a stat instead of a whole-file `JSON.parse`; the conversations list (re-read on every ActivityChanged/ConversationsChanged) now parses only files that changed. Writes stamp the cache; external writers (cloud store-sync hydration, manual edits) invalidate via the stat check — covered by 4 new tests.
3. **Deduped chat-open reads** — the board hydrate + chat-history query fired the same windowed read twice per open; single-flight now shares one promise (wire shows exactly one `?limit=120` per open).

Stable feed-entry keys (the other listed follow-up) already landed in 825d95ed. Gates: Biome clean, all typechecks, boundaries, runtime 747 / web 225 / ui-chat suites green.

cravenceiling:
CI failure fixed in the latest push — it was a test-harness contract gap, not an app bug: the fake host dropped `displayText` on message sends and served content-only histories, so the content-anchored windowed reseed saw a fold differing from the live kickoff bubble and replaced it with the raw setup directive (the reseed churn also broke the Escape-to-close helper). The fake host now persists and serves `displayText` exactly like the real runtime. Full local Playwright run: **133 passed, 1 skipped, 0 flaky**.

PR #1083, opened 2026-07-24
feat: update pi (Earendil) dependencies 0.80.6 → 0.82.0 (HOU-891)

cravenceiling:
Fixes HOU-891.

## What

Updates every `@earendil-works` (pi) dependency from **0.80.6 → 0.82.0** — `pi-ai` and `pi-coding-agent` directly, `pi-agent-core` and `pi-tui` transitively — and migrates Houston across the four breaking-change releases in between (0.80.7, 0.80.8, 0.81.0, 0.82.0).

## Root cause / what broke

pi 0.80.8 removed the exported `AuthStorage`/`ModelRegistry.create` facades and the global OAuth registry (`pi-ai/oauth` is now type-only) in favor of an app-owned `CredentialStore` + async `ModelRuntime`; 0.82 additionally dispatches every stream strictly by *registered provider id*. Houston's runtime used `AuthStorage` as its whole auth facade (sync reads on the turn path, serve-mode `reload()`, OAuth login callbacks), so the bump was a real migration, not a version edit.

## Changes

- **New `HoustonAuthStore`** (`packages/runtime/src/auth/credential-store.ts`): pi-ai `CredentialStore` over the same `dataDir/auth.json` (atomic 0600 writes via the existing `auth-file.ts` helpers), keeping the sync `get/set/reload` surface Houston's turn-time paths and cloud serve mode require. `modify` is serialized per provider so pi's locked OAuth refresh cannot double-refresh a rotated token.
- **`auth/storage.ts`** now builds the singleton `ModelRuntime` (top-level await, offline at boot) over that store, plus the sync `ModelRegistry` facade and a `keyStore` for the usage/balance probes (`getApiKey` → `modelRuntime.getAuth`). All `{authStorage, modelRegistry}` pairs thread `modelRuntime` instead (pi backend, one-shot, titles, anonymize, generate-agent, per-turn cloud sessions).
- **`auth/login.ts`** rewritten onto pi's new `AuthInteraction` protocol: `auth_url`/`device_code` notify events map to Houston's existing `LoginInfo` wire shapes; the Codex login-method select answers by `deviceAuth` (ids `"browser"`/`"device_code"`, pinned locally since pi stopped exporting the constants); Copilot's opening Enterprise-domain text prompt stays auto-answered; `manual_code` prompts resolve from the client-relayed paste promise. Anthropic's sanctioned setup-token flow is untouched.
- **Local (OpenAI-compatible) provider registration**: pi 0.82 streams only registered providers, so every endpoint write now syncs an `openai-compatible` provider registration on the live runtime (`bindCustomProviderRegistrar` at boot), and per-turn cloud runtimes register from their hydrated data dir.
- **Host**: Copilot central refresh reaches pi's flow through `githubCopilotProvider().auth.oauth.refresh` (the standalone export is gone); the cloudrun device-code connect drives `ModelRuntime.login`; the OAuth trio (anthropic / openai-codex / github-copilot) is pinned in `providers/pi-oauth.ts` because pi ≥0.81 gives many api-key providers *optional* OAuth (xai, kimi-coding, openrouter) and neither the catalog wire shape nor the api-key connect gate may flip them.
- **Deleted now-native code** (per each file's own delete-on-bump note): the Bedrock bearer-token bridge (`ai/bedrock.ts` — pi 0.80.7 maps the generic `apiKey` option to a Bedrock bearer token), and both gemini-flash / moonshot-k3 catalog patches + tests in runtime *and* host (Gemini 3.6 Flash, 3.5 Flash-Lite, and Kimi K3 ship natively in 0.82.0).
- Static catalog reads cast to `BuiltinProvider` (the new `KnownProvider` also names the purely-dynamic `radius` provider, which has no static catalog entry).

Known upstream quirk: `ModelRuntime.refresh()` (awaited by `setRuntimeApiKey`) hangs in pi 0.82; nothing on Houston's production paths calls it, and the two tests that did now seed credentials through the store (the production-shaped path).

## Verification

Before the fix (repro): on `main`, edit both `package.json`s to `"@earendil-works/pi-ai": "0.82.0"` / `"@earendil-works/pi-coding-agent": "0.82.0"`, run `pnpm install && pnpm --filter @houston/runtime --filter @houston/host typecheck` → ~57 type errors (`AuthStorage` not exported, `authStorage` not in `CreateAgentSessionOptions`, `pi-ai/oauth` members gone, `radius` id mismatches).

After (this branch):
- `pnpm --filter @houston/host --filter @houston/runtime --filter @houston/domain test` → **1,936 passed** (domain 147, runtime 766, host 1,023)
- `pnpm typecheck` (all 25 workspace projects) → clean
- `pnpm check` (Biome) → clean
- `pnpm check:boundaries` → clean

Functional smoke of the riskiest path: a scripted faux-provider turn streams end-to-end through the new `ModelRuntime` (`summarize.test.ts`), and the local-endpoint turn classification test (`context-overflow.test.ts`) proves a configured custom endpoint dispatches and still yields the typed `context_overflow` error.


cravenceiling:
**Independent review pass (GPT-5.6, high effort) — 2 confirmed findings, both fixed in b3a1f940:**

1. **Login completion was gated on a network catalog refresh.** `ModelRuntime.login()`/`logout()` end with `refresh({allowNetwork: true})` whenever `PI_OFFLINE` is unset, outside the login abort signal — on a stalled network a *successful* sign-in sat unresolved past the 10-minute abandonment expiry (reported "Login timed out", then late-flipped to complete), and the host's device-code connect held its cross-replica lock + a temp credential file for the duration. Both paths now drive the provider's own `auth.oauth.login` and persist via the CredentialStore's documented app-login pattern; the host connect no longer touches a throwaway auth.json at all — the credential goes straight from the in-process login result to the central store.

2. **Sign-out could be resurrected by an in-flight token refresh.** `logout` used the store's sync `remove()`, bypassing the per-provider `modify` chain — a refresh landing afterward re-persisted the rotated credential. `logout` now awaits `authStorage.delete()`, which queues on the same chain. Regression tests added in `credential-store.test.ts` (delete-behind-in-flight-refresh, chain serialization, rejection recovery).

**Cloud-repo impact: none required.** The hosted `/v1/catalog` rides the engine roll (houston's `engine-pod-image.yml` boots the built image and publishes its real catalog as the `engine-pod-catalog:<sha>` artifact; cloud's roll workflow writes it into the gateway ConfigMap), and the Go gateway's mirrored OAuth constants (Codex client id/token URL, Anthropic client id + `platform.claude.com/v1/oauth/token`, Copilot editor headers) are byte-identical to pi 0.82's.

PR #1120, opened 2026-07-28
dev: automated E2E probe for event-driven routines (HOU-893)

cravenceiling:
## What

`scripts/dev/test-triggers.sh` — an automated end-to-end probe that answers "are triggers working?" in one command (HOU-893 asked exactly that and the answer previously required a by-hand gateway walk).

Two modes:

- **Local, hermetic (default).** Spawns a throwaway host from the checkout (`HOUSTON_MANAGED_CLOUD=1`, temp data dir), creates an agent + webhook-trigger routine over the API, POSTs a batch to the internal `/agents/:id/trigger-events` delivery route (what the control plane calls), and asserts exactly one run fires and the event id dedups on redelivery. No docker, no sign-in, no cloud. With `ANTHROPIC_API_KEY` present it seeds the pod-style `credentials.json` so the turn completes for real; without it the script reports the expected no-provider outcome as PASS (pipeline) / WARN (turn).
- **`--staging`.** Full C9 pipeline against the deployed staging gateway: create routine → wait for pod→gateway sync → mint webhook key → trigger-status `active` → POST the public hook URL → assert the run fires. Authenticates from the desktop keychain session (GCIP), refreshed via `FIREBASE_API_KEY` when stale.

## Verification

- Local mode: PASS (pipeline) with the documented no-provider WARN on a credential-less run.
- `--staging`: PASS end-to-end, repeatedly — run fires ~15s after the hook POST with the payload framed as untrusted event data; probe routine cleaned up after each run.

## Notes

- Full-pipeline mode against the `pnpm dev` gateway is documented as blocked: cpdev engines do not store-sync to the control plane, so the trigger projection never runs locally (cloud-repo gap; needs a cpdev fix).
- The script header carries the operational warning to never set `GW_WEBHOOK_PUBLIC_URL` in dev — the local control plane would re-register the shared Composio project's delivery webhook and hijack real trigger deliveries.
- Dev webhook-key minting itself was fixed on main via `GW_PUBLIC_BASE_URL` (#1114), so this branch adds no env changes.

HOU-893

### DIFF
--- a/scripts/dev/test-triggers.sh
+++ b/scripts/dev/test-triggers.sh
@@ -1,0 +1,306 @@
+#!/usr/bin/env bash
+# Automated E2E probe for event-driven routines (C9 triggers).
+#
+# Two modes:
+#
+#   scripts/dev/test-triggers.sh
+#     LOCAL (hermetic, default): spawns a throwaway host from THIS checkout
+#     with HOUSTON_MANAGED_CLOUD=1 and a temp data dir, creates an agent and a
+#     webhook-trigger routine, POSTs a batch to the internal
+#     /agents/:id/trigger-events delivery route (what the control plane calls),
+#     and asserts exactly one run fires + the event id dedups on redelivery.
+#     No docker, no gcloud, no sign-in. This covers the ENTIRE houston half of
+#     triggers (schema, write gate, fire path, dedup lock, prompt framing).
+#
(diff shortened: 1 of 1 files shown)


cravenceiling:
Closing — keeping trigger verification as throwaway scripts instead of committed tooling.

PR #1128, opened 2026-07-28
feat(skills): global Skills manager in the sidebar (HOU-792)

cravenceiling:
### Root cause / finding

The issue's premise ("custom skills are saved at the org level rather than the agent level") turned out to be **factually inverted**: every skill path in the codebase — CRUD routes, the HOU-791 skill builder, marketplace installs, the pi runtime loader — resolves strictly to the per-agent `<agent>/.agents/skills/` directory. There is no org-level skills storage anywhere, so skills were never leaking between agents.

What WAS missing is the other half of that observation: there was no way to manage skills across agents at all. Per product direction, this PR embraces cross-agent skills with a global **Skills** section in the left sidebar.

### What this adds

A new top-level **Skills** page (viewMode `skills-home`, sidebar entry between Integrations and AI Models) — a pure client-side aggregation over the **existing per-agent routes**. The host already authorizes skill operations for any agent the caller owns (`canUseAgent` = workspace ownership), so **zero backend changes** were needed — which also keeps it working against the hosted gateway, which only proxies `/agents/:slug/*`.

- **Your skills** — one row per slug aggregated across the workspace's agents, with an avatar stack showing which agents hold it. Queries reuse the per-agent `queryKeys.skills(path)` keys, so `SkillsChanged` events keep the page live for free; fetched once per mount, never on focus (the `useAllConversations` pod-waking discipline).
- **Manage dialog** (row click) — edit the full SKILL.md and toggle which agents hold the skill. `planSkillAssignment` diffs the edit into a minimal fan-out: newly assigned agents always receive the canonical content; existing holders are rewritten only when the content was actually edited (an assignment-only save never clobbers a divergent per-agent copy); unassignments require an explicit confirm naming the agents; Delete removes it from every holder behind a confirm.
- **skills.sh Store tab** — the same `SkillMarketplaceSection` as the per-agent tab; installing opens a pick-agents dialog (agents that already have the skill are locked out) and fans out per-agent installs, with per-failure toasts naming the agents.
- **New skill** — header CTA with the from-scratch fields plus an agent multi-select; creates on every picked agent. Agents already holding the derived slug lock out live as you type.

Pure models (`aggregateWorkspaceSkills`, `filterWorkspaceSkills`, `planSkillAssignment`) live in `app/src/lib/workspace-skills.ts`, node:test-covered. i18n shipped for en/es/pt. KB updated (`knowledge-base/skills.md`).

Note on visual baselines: the new sidebar row lands inside the visual suite's committed 1% `maxDiffPixelRatio`, so the board baselines still pass unchanged and were deliberately not re-recorded.

### Repro (before)

1. Open Houston. There is no global place to see or manage skills — the sidebar has no Skills section.
2. To give three agents the same skill you must open each agent's Agent Settings → Skills and install/create it three times; there is no way to copy or unassign a skill across agents.

### Verify (after)

1. `pnpm dev`, open the app: the sidebar shows **Skills** between Integrations and AI Models.
2. The page lists every skill across the workspace's agents, each row showing the holder agents' avatars; search matches title, slug, and agent name.
3. Install a store skill: the pick-agents dialog appears; confirm installs it to each picked agent (check each agent's own Skills tab shows it).
4. Click a row: edit the content and toggle an agent on → Save writes the copy to that agent. Toggle an agent off → Save asks for confirmation naming the agent, then removes only that copy.
5. **New skill** creates the skill on every selected agent.
6. Gates: `pnpm check`, app `tsgo --noEmit` + `pnpm test` (2270 pass), `pnpm --filter houston-web test:e2e` (165 pass), `test:visual` (6 pass), `check:parity`, `check:boundaries`, `check-locales` — all green.

### DIFF
--- a/app/src/components/shell/sidebar-chrome.tsx
+++ b/app/src/components/shell/sidebar-chrome.tsx
@@ -1,11 +1,19 @@
 import type { SidebarLabels, SidebarNavItemEntry } from "@houston-ai/layout";
 import { WorkspaceSwitcher } from "@houston-ai/layout";
 import type { TFunction } from "i18next";
-import { Blocks, Boxes, LayoutDashboard, Settings, Store } from "lucide-react";
+import {
+  Blocks,
+  Boxes,
+  LayoutDashboard,
+  LibraryBig,
+  Settings,
+  Store,
+} from "lucide-react";
 import { useState } from "react";
 import { useCapabilities } from "../../hooks/use-capabilities";
 import { hasSpaces } from "../../lib/org-roles";
 import { INTEGRATIONS_VIEW_ID } from "../integrations-view";
+import { SKILLS_VIEW_ID } from "../skills-view/id";
 import { STORE_VIEW_ID } from "../store-view";
 import { CreateTeamDialog } from "./create-team-dialog";
 
@@ -40,6 +48,13 @@ export function buildSidebarNavItems(args: {
       icon: <Blocks className="h-4 w-4" />,
       onClick: () => setViewMode(INTEGRATIONS_VIEW_ID),
       dataAttrs: { "data-tour-target": "nav-integrations" },
+    },
+    {
+      id: SKILLS_VIEW_ID,
+      label: t("shell:sidebar.skills"),
+      icon: <LibraryBig className="h-4 w-4" />,
+      onClick: () => setViewMode(SKILLS_VIEW_ID),
+      dataAttrs: { "data-tour-target": "nav-skills" },
     },
     ...(showAiModels
       ? [
--- a/app/src/components/shell/top-level-screen-views.tsx
+++ b/app/src/components/shell/top-level-screen-views.tsx
@@ -7,6 +7,7 @@ import { AiHubView } from "../ai-hub/ai-hub-view";
 import { Dashboard } from "../dashboard";
 import { INTEGRATIONS_VIEW_ID, IntegrationsView } from "../integrations-view";
 import { SettingsView } from "../settings/settings-view";
+import { SKILLS_VIEW_ID, SkillsView } from "../skills-view";
 import { STORE_VIEW_ID, StoreView } from "../store-view";
 import type { KeepAliveView } from "./keep-alive-views";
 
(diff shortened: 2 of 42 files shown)


cravenceiling:
## Round 2 (product feedback)

- **Setup chats are now side-by-side (the Routines split)** on both Skills surfaces: `SkillSetupChat` portals into the shell's right-hand detail panel and owns the panel-open flag on mount/unmount, so the catalog stays visible on the left while the conversation runs on the right (Escape closes, routines parity). The old swap-the-catalog behavior is gone.
- **Global page: Custom skills tab** — Create skill (guided chat; a multi-agent workspace first picks who builds it, one-agent workspaces go straight to the chat), Add skill (the multi-agent from-scratch dialog), and the Houston library shelves. Library installs go through the same pick-agents fan-out as marketplace installs (one unified pending-install flow). The "New skill" header CTA now starts the chat.
- **Per-agent Custom tab: "From your other agents"** — the user's own skills living on other agents in the workspace, one click copies the SKILL.md verbatim onto this agent (rows show holder avatars; already-installed rows show a quiet check). Mounted only inside the tab content so the cross-agent fetch runs only when the tab opens.

Gates rerun: Biome, workspace typecheck, 2270 unit tests, 165 e2e, visual suite — all green. Screenshot-verified both splits (global page + agent settings) against the fake host.

cravenceiling:
Round 3: the manage dialog gains an **Edit in chat** button (closes the dialog, opens the skill's guided chat in the side panel — on the global page hosted on the skill's first holder). The per-agent Skills tab now opens this **same dialog** on row click instead of jumping straight into the chat; the chat header's Edit-manually lands on the dialog too. The dialog's holder aggregation is fetched lazily only while it's open, and on an agent's tab that agent is pinned first so ITS copy is the canonical content. Read-only surfaces keep the raw edit modal. Gates: typecheck, Biome, 2270 unit tests, 165 e2e — all green.

cravenceiling:
Round 4: 'From your other agents' rows now open the shared skill preview modal (same as store/library skills): title, 'From the {agent} agent' by-line, description, connected apps, category, and the full instructions behind the expander. The body loads on open from the holder agent (cached useSkillDetail); the modal's Install is the one commit point — the row and its + both preview, per the library convention. Gates green (typecheck, Biome, 2270 unit, 165 e2e).

PR #1133, opened 2026-07-28
perf(search): stream bounded mission-search scans instead of full-transcript waves

ja-818:
> Stacked on #1130 — merge that first; this PR auto-retargets to `main` once its base branch is deleted (or retarget manually).

## Summary
Mission search sat on a loading state for 10-15+ seconds. Four compounding causes, all fixed:
- **No debounce**: every keystroke launched a new wave of history fetches that was never canceled. Now the network wave debounces 250 ms and a superseded phrase stops the running wave from starting more work. Matching what is already loaded stays per-keystroke (it is now cheap).
- **Full transcripts**: bulk search reads fetched each mission's ENTIRE transcript (chat opens are windowed to 120 messages per HOU-819; search was not). Bulk scans now read a bounded `SEARCH_SCAN_WINDOW` (400 messages) and skip the IndexedDB conversation-cache write (a partial transcript must never be cached as complete, and each write enumerated + pruned the whole store).
- **~100× slow folding**: `foldForSearch` walked text char-by-char with per-char `normalize()` on full transcripts, per keystroke. It is now a whole-string fold; transcripts are folded ONCE at load into `SearchableText` (original + folded), so per-keystroke matching is a regex test. The char-by-char fold map survives only for highlight ranges on short snippet text.
- **All-at-once commit**: results waited on `Promise.allSettled` of every mission. A bounded pool (5 concurrent) now streams results onto the board as transcripts land, coalesced in 120 ms commit windows.

## Testing
- New unit tests: `app/tests/mission-search-text.test.ts` (fold semantics incl. accents/NFKD, snippet extraction on originals), `app/tests/async-pool.test.ts` (bounded concurrency, stop condition).
- `cd app && pnpm test`: 2276 pass. `pnpm --filter houston-web typecheck` + `typecheck:e2e` clean. Biome clean.
- Playwright `board.spec.ts`: search test and suite pass (one unrelated test flaked identically on baseline under machine load, verified by A/B with changes stashed).

## Not done (follow-up)
- `packages/sdk/src/modules/missions-search/search-text.ts` (headless port for native surfaces) still carries the old per-char fold; not on the desktop/web path.
- A server-side search endpoint (match inside conversation files in the runtime, return ids + snippets) would also eliminate the cold-runtime-spawn cost that search triggers per agent; out of scope here.

Fixes HOU-941

### DIFF
--- a/app/src/components/board/board-source.ts
+++ b/app/src/components/board/board-source.ts
@@ -118,7 +118,8 @@ export interface BoardSource {
   /**
    * Persisted chat history for one conversation. Callers forward `opts`:
    * mission search bulk-loads with `observe: false` (no per-conversation
-   * observer streams); the board's open-a-chat hydration omits it (observes).
+   * observer streams, bounded scan window); the board's open-a-chat hydration
+   * omits it (observes, tail window).
    */
   loadHistory: (
     sessionKey: string,
--- a/app/src/components/mission-highlight.ts
+++ b/app/src/components/mission-highlight.ts
@@ -15,9 +15,21 @@ function foldChar(char: string): string {
   return char.normalize("NFKD").replace(COMBINING_MARKS, "").toLowerCase();
 }
 
-/** Fold `text` for matching while recording, for each folded char, the index of
- *  the original char it came from. `map[folded.length] === text.length`, so an
- *  exclusive end position always resolves. */
+/**
+ * Fold `text` for matching while recording, for each folded char, the index of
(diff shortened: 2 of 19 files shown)


ja-818:
Added a follow-up commit from the adversarial review pass: a debounced phrase change could supersede a scan wave whose claims were released only after the new wave had skipped them, leaving most transcripts permanently unscanned while the search reported complete. The wave bookkeeping is now a pure scanner module with unit tests pinning that scenario.

PR #1138, opened 2026-07-28
fix(board): never let a login race or one sick agent blank the mission board

ja-818:
> Stacked on #1133 (which stacks on #1130). Merge order: #1130 → #1133 → this.

## Summary
Intermittently the mission board was empty (or a day old) right after login. Root causes and fixes:

- **Frozen restored cache**: the `all-conversations` aggregate used `staleTime: Infinity`, and IndexedDB persistence restores it with its original `dataUpdatedAt` — so a restored board was never revalidated all session, hiding every mission created while the app was closed (overnight routines, teammates, other devices). Now a finite 10-minute freshness window: the boot that restores the cache revalidates it, mid-session navigation stays cached, and no background sweep wakes the pod fleet. (Deliberately NOT `refetchOnMount: "always"` — seven surfaces mount this hook, and per-mount refetches would fan out to every agent's pod on each navigation.)
- **One sick agent blanked everything**: the per-agent fan-out used `Promise.all`, so a single failing read (e.g. a pod cold start outliving the gateway's ~2s of retries) rejected the sweep; with `retry: false` the error was cached, `placeholderData` doesn't apply on error, and `isLoaded: isFetched` rendered the empty board with the new-mission composer auto-opened. The sweep is now partial-tolerant (`allSettled`; agents that answered render, failed agents keep their last-known rows), reports which agents failed (one branded toast + a bounded self-rescheduling re-sweep at 5s/20s/60s), transient failures get a bounded retry, and `isLoaded` gates on success — an error paints the last-known missions, never a fake-empty board.
- **SSE catch-up seam never wired**: `global-events`' documented `onConnect` reconnect seam is now passed through the adapter; a stream reconnect invalidates the aggregate so events missed during the gap are recovered.
- **Silent workspace fallback**: `listWorkspaces`' blanket `catch { return [personal] }` silently dropped all `org:*` spaces on a transient failure, stranding Teams users in the personal space with all missions "missing". Now: bounded transient retry, 404 still degrades (capability negotiation), anything else throws into the existing visible failed state + retry, and the persisted `last_workspace_id` survives for the next successful load.

New contract: `listAllConversations` returns `{ conversations, failedAgentPaths }` so partial sweeps are explicit, never silent.

## Testing
- New unit tests: partial-sweep merge + recovery plan, transient-error classifier, invalidation-plan reconnect case, adapter partial-tolerance, workspace degrade/throw + preference survival, events reconnect catch-up.
- New Playwright cases in `board.spec.ts` (via a new fake-host `fail-agent-reads` hook): one agent 500s → healthy agents' missions still render + toast; aged restored mirror with an offline-written mission and `/v1/events` blocked → the boot sweep brings it in.
- `pnpm check` (full biome), `check:boundaries`, app `tsgo` + 2292 unit tests + `check-locales`, houston-web typechecks + 282 unit tests, full Playwright chromium suite (161 passed) — all green.
- KB updated: `knowledge-base/files-first.md` cross-agent aggregate contract.

Fixes HOU-981


ja-818:
Added a follow-up commit from the adversarial review pass: the empty-board verdict now waits out the placeholder paint; the sweep retry surfaces exactly one toast/capture per outage; the reconnect catch-up survives token-refresh resubscriptions; and partial-sweep recovery state is roster-keyed so a space switch resets it.

PR #1161, opened 2026-07-30
fix(chat): keep the mission log live when tools follow streamed narration (HOU-1047)

cravenceiling:
## Summary

Fixes HOU-1047.

**Symptom:** when an OpenAI-backed agent does async work (e.g. connecting to anakin.io, browsing the web), the chat shows the generic loading dot instead of live activity — and the tool usage appears stranded inside the previous message's (settled) mission log instead of a live one at the bottom.

**Root cause:** OpenAI-style turns narrate BEFORE running tools ("Let me connect to anakin.io…" → then the tool calls). The streamed narration keeps the current assistant message open in `feedItemsToMessages`, so the first `tool_call` fused into the text message. The grouping layer (`getChatDisplayItems`) flushes a process block that shares a message with content as **inactive, above the content** — so no trailing active mission log existed, and `shouldShowThinkingIndicator` fell back to the generic loading indicator. Claude-style turns (thinking → tools → text) never hit this because no text precedes the tools; the `thinking` case has always flushed a content-bearing message, but `tool_call` did not.

**Fix:** in `ui/chat/src/feed-to-messages.ts`, a `tool_call` that arrives while the current assistant message already has visible content flushes it and starts a fresh process-only message — mirroring the existing thinking-case flush. The tools become the trailing process block, which renders as the **active mission log at the bottom** (current action in the header, e.g. "Anakin · Executing") and suppresses the standalone loading indicator. Placeholder-input dedup and `tool_result` matching carry across the flush boundary unchanged (covered by tests).

## Reproduction (before the fix)

1. `pnpm dev`, open an agent whose chat provider is an OpenAI model (Codex OAuth or API key).
2. Ask for something that makes the model narrate and then run tools, e.g. "Connect to anakin.io and fetch my account details" (any integration/browse flow works).
3. The model streams a sentence like "Let me connect to anakin.io…", then runs tools: the bottom of the chat shows only the pulsing loading indicator; expanding the chevron on that last text bubble reveals the tool calls hidden in a settled "Mission log" above the text.

## Verification (after the fix)

1. Same flow as above.
2. While the tools run, the bottom of the chat shows a live mission-log line (per-tool icon/brand + the action verb, animated), no generic loading indicator, and the chevron expands the streaming log in place.
3. When the turn settles, the log header reads "Mission log" under the narration text; a reload renders the turn from history unchanged (that path already ordered tools before text).
4. `pnpm --filter @houston-ai/chat test` — 320 tests pass, including the new `feed-to-messages.test.ts` suite covering the split, the active trailing block, key stability (HOU-717 class), and dedup/result matching across the flush.

### DIFF
--- a/app/src/components/use-chat-display-labels.tsx
+++ b/app/src/components/use-chat-display-labels.tsx
@@ -13,13 +13,23 @@ export function useChatDisplayLabels(): Pick<
   // tense label the process header shows as a branded row; ui/chat calls it
   // through `processLabels.resolveActionBrand`, staying Composio-unaware.
   const resolveActionBrand = useActionBrandResolver();
+  // The same astronaut deck the pre-reply indicator plays. Passed into
+  // processLabels too so the ACTIVE mission-log header keeps the personality
+  // through mid-turn reasoning gaps — the standalone indicator below is
+  // suppressed the whole time an active log trails (HOU-471/HOU-1047), so
+  // without this the header held the last tool's stale verb.
+  const loadingPhrases = useMemo(
+    () => t("loadingPhrases", { returnObjects: true }) as string[],
+    [t],
+  );
   const processLabels = useMemo(
     () => ({
       active: t("process.active"),
       complete: t("process.complete"),
+      phrases: loadingPhrases,
       resolveActionBrand,
     }),
-    [t, resolveActionBrand],
+    [t, loadingPhrases, resolveActionBrand],
   );
   const getThinkingMessage = useCallback<
     NonNullable<ChatPanelProps["getThinkingMessage"]>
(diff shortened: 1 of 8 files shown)


cravenceiling:
Verified the fix against production screenshots of the bug (OpenAI agent connecting anakin.io). All three visible symptoms trace to the same root cause this PR fixes:

1. **Static "Mission log" above the narration while the agent works** — prod fuses the tool calls into the streamed-text message, so the grouping layer flushes the block *inactive* (static label, no shimmer). Post-fix the tools form a trailing block flushed *active*: the header shimmers and names the current action.
2. **The cycling status phrase at the bottom** — it only shows because no active process block trails the conversation; post-fix `shouldShowThinkingIndicator` suppresses it while the live log is on screen.
3. **A "new" mission log appearing below the text after completion** — at settle, the final `assistant_text` finalizes the text entry (positioned before the tools in the feed), which re-groups the tools into their own message for the first time: the log teleports from above the text to below it. Post-fix the trailing block exists throughout the run and settles in place — same React key, active → inactive, no remount and no jump (new test).

Also added a test for the multi-phase shape seen in the fourth screenshot (tools → narration → more tools): the first phase's log settles above the text and the second phase starts a fresh live trailing log, instead of silently growing the settled one.

cravenceiling:
Follow-up from testing: with the trailing log now correctly active, the standalone rotating-phrase indicator (HOU-910) is suppressed for the whole turn (HOU-471: one in-flight signal), so the astronaut one-liners disappeared mid-turn and the header held the last tool's stale verb through every reasoning gap.

New commit: the active header now plays the same localized phrase deck whenever nothing is visibly executing — before the first tool, and once the latest tool has its result. A genuinely running tool still shows its verb or the integration brand row. Without a deck passed in (ui/chat standalone) the sticky-verb behavior (HOU-448) is unchanged. The rotation hook is shared with ChatThinkingIndicator and keyed off the block's active flag so a quick tool flashing its verb doesn't reshuffle the deck.

PR #1176, opened 2026-07-31
Per-mission models, connected-provider defaults, credential single-rotator + self-heal, Ask First missions

ja-818:
## What this ships

**Per-mission models (HOU-1064).** A model pick inside an open mission pins that mission only — including multiplayer Teams mode, where a pick previously moved the per-(user, agent) choice shared by every conversation. The picker displays the mission's in-ceiling pin, every send forwards the displayed pin, ceiling snaps resolve the model's owning provider, displayed effort is revalidated per model, and personal choices update optimistically. Companion gateway change: gethouston/cloud PR (linked below) — the gateway now honors an in-ceiling per-conversation pin verbatim.

**Connected-provider defaults (HOU-1065).** The create-agent dialog and import wizard no longer hardcode anthropic/Sonnet: defaults resolve against confirmed-connected providers at open and at submit time, and when nothing is confirmed connected the kickoff carries no pin at all — the engine resolves a connected provider or surfaces the real connect flow. Onboarding writes any connected provider into the assistant's config, and the sticky last-used preference can no longer be silently lost (hosted picks now update it too).

**Credential single-rotator + self-heal (HOU-1073).** Root causes of "every new agent's first message is broken": the connect flow materialized full refresh-bearing credentials onto pods (whose SDK rotation burned the gateway's central copy), and the connect flow's own status polling clobbered fresh logins before capture could export them. Now: serve-apply never overwrites a refresh-bearing local entry, capture retries and reports honest failure, the scrub is coupled to capture, serve-mode pods materialize Anthropic access-only, a pinned turn on a definitively absent provider yields the real connect card instead of an SDK 401, a serve miss self-heals by re-capturing a warm pod's live credential (single-flight + cooldown), and the warming kickoff verifies its pin against the new pod — dropping the pin and the mission's stamp when the provider is absent (ActivityUpdate provider/model are now nullable to support clearing).

**Ask First missions (HOU-1080).** Every new mission opens in Ask First: the default turn mode is execute and the per-agent remembered mode is removed; a mid-mission pick applies to that session (and live turn) only.

## Verification
- app node:test 2536/2536 · runtime 881 · host 1141 · domain 181 · protocol 29 · web 301 — all green post-rebase onto main
- workspace typecheck, biome, locales (en/es/pt), `check:boundaries` green
- Adversarially reviewed (two independent review passes); all confirmed findings fixed
- Chat/board Playwright specs green; the create-agent-dialog specs could not run in the CI-less sandbox (vite warm-up timeout) — flagged for in-app review

## Ops note
Orgs whose central Anthropic/OpenAI credential is already burned need one reconnect after release; the self-heal also repairs the row automatically from any warm pod that still holds a live credential.

Sibling PR: gethouston/cloud (gateway pin honoring) — link in first comment.

Fixes HOU-1064
Fixes HOU-1065
Fixes HOU-1073
Fixes HOU-1080

### DIFF
--- a/app/src/components/board/use-mission-control-archived-send.ts
+++ b/app/src/components/board/use-mission-control-archived-send.ts
@@ -5,8 +5,8 @@ import { useTranslation } from "react-i18next";
 import { analytics } from "../../lib/analytics";
 import { buildAttachmentPrompt } from "../../lib/attachment-message";
 import { classifyFileKind } from "../../lib/file-kind";
-import { tauriAttachments, tauriChat, tauriConfig } from "../../lib/tauri";
-import { readAgentTurnMode } from "../../lib/turn-mode";
+import { tauriAttachments, tauriChat } from "../../lib/tauri";
+import { DEFAULT_TURN_MODE } from "../../lib/turn-mode";
 import type { Agent, AgentDefinition } from "../../lib/types";
 import { useUIStore } from "../../stores/ui";
 
@@ -62,7 +62,7 @@ export function useMissionControlArchivedSend({
           mode: mode?.promptFile,
           providerOverride,
           modelOverride,
-          modeOverride: await readAgentTurnMode(agentPath, tauriConfig.read),
+          modeOverride: DEFAULT_TURN_MODE,
           mentions,
         });
         analytics.track("chat_message_sent", {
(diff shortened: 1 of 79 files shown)


ja-818:
Sibling gateway PR: gethouston/cloud#207 (dispatch clamp honors in-ceiling per-conversation pins). Land this PR first, then the cloud one.

PR #1180, opened 2026-07-31
feat(providers): qwen extension provider — Alibaba Model Studio pay-as-you-go (HOU-1077)

cravenceiling:
Follow-up to #1178 (HOU-1077). That PR fixed the reconnect dead-end and the connect guidance, but the underlying gap remained: pi-ai ships only the Qwen **Token Plan** gateways, whose endpoint rejects every key except the dedicated one minted after purchasing a plan — the regular (free-quota) Alibaba Model Studio keys users actually create had no provider to connect to. That is exactly what the affected user hit ("Qwen said no paid plan was needed" — true for the standard API, which Houston didn't offer).

## What this adds

A Houston-owned `qwen` provider on Alibaba Model Studio's **international pay-as-you-go endpoint** (`dashscope-intl.aliyuncs.com/compatible-mode/v1`, OpenAI-compatible). Regular Model Studio keys connect with the normal paste flow; no plan purchase needed.

## How (no monkey-patch)

pi-ai has no registry hook for a new provider id (its `MODELS` table is not exported), so the extension threads through Houston's own catalog seams:

- **Runtime** — `qwen-dashscope.ts` derives the model list from pi's own `qwen-token-plan` table (same Model Studio catalog; provider id + base URL remapped, `qwen3.7-max` pinned first as default), so a pi bump updates it in lockstep and compat/thinking flags can never drift. `pi-catalog.ts` lists it, `safeGetModel` resolves it, and both the long-lived and per-turn ModelRuntimes register it for stream dispatch + stored-key auth — the same `registerProvider` mechanism the local OpenAI-compatible provider uses.
- **Host** — twin module (same pattern as the retired moonshot-k3 catalog-patch twins; host must not import `@houston/runtime`): `/v1/catalog` advertises it, the api-key connect gate accepts it, the capabilities hint names it.
- **Frontend** — `qwen` override (name, Model Studio console key link, "free quota, then pay as you go"); the existing `qwen` brand mark and color apply directly. The overrides drift guard now validates against the shipped catalog (pi + this extension).

Both twins carry a DELETE-WHEN-PI-SHIPS note for the day pi adds a DashScope provider natively.

## Verification

- **Live smoke (real network):** the runtime's `verifyApiKey("qwen", <bad key>)` resolved `qwen3.7-max`, dialed DashScope, got the real `401 invalid_api_key`, and classified it `invalid_key` — the exact chain a real key rides, proving endpoint, model dialect, request shape, and classification end-to-end. A valid-key chat test needs a Model Studio account (I don't hold a key); the turn path uses the same registration mechanism the local provider ships on.
- Runtime vitest 889 ✓ (6 new), host vitest 1144 ✓ (3 new), app node:test 2546 ✓, full workspace typecheck ✓, biome ✓, boundaries ✓.

## Verify in app

1. Settings → search "qwen" → two cards: **Qwen** (pay as you go) and **Qwen Token Plan** (subscription).
2. Connect Qwen with a regular Model Studio key ("Get your API key" opens the console) → card flips connected, default model Qwen3.7 Max.
3. Chat; a bad key shows "Qwen didn't accept this key" at connect, and mid-chat auth failures render the reconnect card whose button opens the key dialog (#1178).

### DIFF
--- a/app/src/components/shell/provider-logo-map.ts
+++ b/app/src/components/shell/provider-logo-map.ts
@@ -110,6 +110,9 @@ export const BRAND_ALIASES: Readonly<Record<string, BrandKey>> = {
   "xiaomi-token-plan-ams": "xiaomi",
   "xiaomi-token-plan-cn": "xiaomi",
   "xiaomi-token-plan-sgp": "xiaomi",
+  "qwen-token-plan": "qwen",
+  "qwen-token-plan-cn": "qwen",
+  "qwen-token-plan-individual": "qwen",
   // AI-hub lab ids (see `catalog-lab.ts`) that differ from the provider id.
   // Most lab ids ARE provider ids (anthropic, openai, mistral, deepseek, xai,
   // minimax, zai, nvidia, meta, qwen, cohere, ...) so `providerBrandKey`
--- a/app/src/lib/provider-overrides.ts
+++ b/app/src/lib/provider-overrides.ts
@@ -652,6 +652,20 @@ export const PROVIDER_OVERRIDES: Record<string, ProviderOverride> = {
     // (`provider-api-key-guide.tsx`).
     apiKeyUrl: "https://org.ngc.nvidia.com/setup/api-keys",
   },
+  // Houston's qwen extension provider (not a pi builtin — the host appends it
+  // to /v1/catalog, see packages/host/src/providers/qwen-dashscope.ts): Qwen
+  // on Alibaba Model Studio's INTERNATIONAL pay-as-you-go endpoint, the home
+  // for regular (free-quota) Model Studio keys the Token Plan card rejects.
+  qwen: {
+    name: "Qwen",
+    subtitle: "Alibaba Model Studio",
+    description: "Qwen models from Alibaba, pay as you go.",
+    cost: "Free quota, then pay as you go",
+    installUrl: "https://modelstudio.console.alibabacloud.com",
+    apiKeyUrl:
+      "https://modelstudio.console.alibabacloud.com/?tab=playground#/api-key",
+    defaultModel: "qwen3.7-max",
+  },
   // Alibaba's prepaid token bundles for Qwen (+ hosted open models). The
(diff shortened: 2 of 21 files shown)


cravenceiling:
Rebased onto main's pi 0.84.1 bump and verified against it:

- **No collision, still needed**: 0.84.1 adds `qwen-token-plan-individual` — the personal tier of the SAME token-plan endpoint (dedicated purchased-plan keys only). pi still ships no pay-as-you-go DashScope provider, which is what this PR adds; there is no `qwen` id in pi's registry.
- The derived model list absorbed 0.84.1's catalog changes automatically (`qwen3.8-max` replacing the preview id) — exactly the lockstep the derivation was built for.
- Polish folded in for the new card: a `qwen-token-plan-individual` override (name, personal-plan quickstart link, subscription cost line, Qwen default model) and real Qwen brand marks for all three token-plan variants (they rendered a bare "Q" monogram before).

Gates on 0.84.1: runtime 1029 ✓, host 1338 ✓, app 2769 ✓, workspace typecheck ✓, biome ✓.

cravenceiling:
Live testing surfaced a real gap, now fixed: **Model Studio international keys are region-scoped.** A key created in US (Virginia) answers `invalid_api_key` on the Singapore endpoint (`dashscope-intl`) and authenticates on `dashscope-us` — verified with a real key both ways (a Singapore key connects and chats end-to-end ✅).

The verify probe now tries each region and persists the accepting one (`qwen-region.json` beside settings.json, read by every model resolution path incl. the per-turn hydrated dir) — mirroring `nvidia-verify`'s shape: that retries models on one endpoint, this retries endpoints with one model. Also classifies DashScope's `AllocationQuota.FreeTierOnly` body (valid key, exhausted free tier) as `quota_exhausted` instead of a rate-limit countdown, and the three Qwen cards now say exactly which Alibaba key each accepts.

Gates: runtime 1039 ✓ (region + verify tests), host 1338 ✓, app 2769 ✓, typecheck ✓, biome ✓.

PR #1182, opened 2026-07-31
fix(integrations): custom POSTs survive mixed undici stacks, manual add returns, deterministic discovery (HOU-1083)

cravenceiling:
Fixes HOU-1083.

### 1. Every custom-integration POST failed: `invalid content-length header (UND_ERR_INVALID_ARG)`

Effect's HttpClient (`@executor-js` engine internals) sets an explicit `content-length` header on every sized request body. In the host process, pi installs npm undici 8.5's fetch + global dispatcher over Node's bundled undici; in that mixed stack the header reaches the wire **twice**, and undici v8's validator rejects the merged value (`"38, 38"`). GETs carry no body, so reads worked while every write died client-side, 100% of the time. Root-caused **live on an engine pod** with a CDP breakpoint inside undici's `processHeader` (captured `val = "38, 38"`).

Fix: the executor now owns its HTTP seam (`packages/host/src/integrations/custom/fetch-guard.ts`) — a guarded `httpClientLayer` + `fetch` passed to `createExecutor` strips message-framing/hop-by-hop headers (`content-length`, `transfer-encoding`, `connection`, `host`, ...) before delegating to the *current* `globalThis.fetch`, so whichever fetch implementation is live computes the single correct value itself. Deterministic on any Node/Bun/undici mix; covers OpenAPI and MCP (the MCP transport derives its fetch from the same layer).

Upstream bug found on the way: `@executor-js/sdk`'s promise-surface `createExecutor` **silently drops** `httpClientLayer`/`fetch` from its config allowlist (runtime and type). `patches/@executor-js__sdk@1.5.37.patch` forwards them; worth an upstream PR. Executor bumped 1.5.32 → 1.5.37 (latest dist-tag; same effect pin, additive API).

`fetch-guard.test.ts` recreates the hostile duplicating pipeline and runs a real executor POST through it — it fails with the verbatim production error if either the guard wiring or the sdk patch is removed.

### 2. Manual "Add custom integration" was gone

PR #1171 deleted the manual form outright. Restored as a fork dialog — "Set up with your agent" (lead) / "Add manually" — composed with #1171's per-agent transport (`useCustomTransportAgentId`, gateway-safe) and the HOU-980 review hardening (URL-keyed detect race guard, cache seeding on add, aria-pressed kind toggle, AsyncButton submits). en/es/pt strings, node:test + Playwright specs restored.

### 3. Non-deterministic action discovery (19 vs 5 vs 0 actions)

The executor compiles specs deterministically (same spec → same actions, verified repeatedly with the production anakin spec: always 19). The variance came from the *agent authoring a different spec each run* and having **no repair path** (duplicate slug → 409), which bred second integrations like `anakin_io_search`.

- Product prompt (TS host + Rust shell twin): fixed discovery order (published OpenAPI doc → `llms-full.txt`/`llms.txt` → docs pages), complete-coverage spec authorship (every documented operation), and a closed verification loop against the compiled action count.
- New `replace: true` on add (manager + sandbox route + agent tool): in-place spec swap keeping the stored credential (no key re-entry) and `addedAtMs`; a replacement that fails to compile restores the previous working view and rejects.

## Verification

- Before (prod, staging pod `houston-o-…`): `POST /v1/integrations/custom/execute` on the live host → `Transport error (POST https://api.anakin.io/v1/search): fetch failed: invalid content-length header (UND_ERR_INVALID_ARG)`; same call through the standalone patched pipeline → clean 401/200.
- `pnpm --filter @houston/host --filter @houston/runtime --filter @houston/domain test` — 2214 passed.
- `cd app && pnpm tsgo --noEmit && pnpm check-locales && pnpm test`; `cargo check` (shell); `pnpm --filter houston-web typecheck`; `pnpm check:boundaries`; Biome clean.
- Playwright: `custom-integrations.spec.ts` + `integration-setup-chat.spec.ts` — 12/12 headless.

After deploy, re-test in-app: execute an anakin.io POST action (search / wire task) — it should return data; then delete the duplicate `anakin_io_search` and ask the agent to fix `anakin_io` — it should replace in place, not create a third.


cravenceiling:
**Independent review** (Codex, gpt-5.6 high effort) ran against the full diff. Six findings — five fixed in 71979914, one was a prompt overclaim (reworded):

1. **Blocker — credential exfiltration via `replace:true`**: a replacement could carry the stored key to a spec whose `servers[].url` points anywhere. Fixed: the carry now requires proof of an unchanged service (`service-origins.ts` — equal server/endpoint origins; url-kind specs must keep the identical document URL; indeterminate = changed). A moved origin drops the carry, deletes the old secret, lands the def `pending` for secure-card re-entry.
2. **Major — non-transactional restore**: a replace failing after `addSpec` left the partial compile occupying the slug. Fixed: uncompile before restoring.
3. **Major — unsynchronized mutations**: add/replace/setCredential/remove now serialize through one manager queue (concurrent replace+remove test).
4. **Major — orphaned secrets** on credential→none replaces: deleted while the old def still names them.
5. **Major — OAuth-only MCP**: manual form disables Add on an OAuth verdict; host probe now errors honestly instead of reporting "active, 0 actions".
6. **Minor — prompt overclaim**: pending adds report no action count; both prompt copies route the completeness check through `integration_search` after key save.

Reviewer confirmed clean: fetch-guard header selection, the sdk patch's seam coverage, manual-form race handling, per-agent routing, locale structure.

New coverage: `service-origins.test.ts` (7), five manager scenario tests, `oauthBlocked` node test. All gates re-run green (host 1162, app 2566, cargo, web typecheck, Biome).

cravenceiling:
Second independent review round triaged (it reviewed the pre-hardening commit; the stranded-secret half of its main finding was already fixed in 71979914). Two findings survived, both addressed in b013a853:

1. **Silent auth downgrade via replace**: a same-service replace replaying `auth:"none"` discarded a working credential, contradicting the tool's "key survives" promise (the caller is a model re-deriving `auth` every call). Replace is now spec-repair proper: existing credential + unchanged origins = key kept and `auth` forced back to `credential`, whatever the input said. A moved origin still drops + deletes (the exfiltration guard from round one wins).
2. **`guardedFetch` Request-object inputs**: headers carried on a `Request` input bypassed the strip (unreachable with today's callers, but the seam's claim should hold unconditionally). Cloned with sanitized headers; test proves the unread body survives the clone.

Suite: 73 passed / 1 skipped in the custom-integration package; host typecheck + Biome clean.

PR #1229, opened 2026-08-04
fix(ci): staging builds stop writing to the production Agent Store

cravenceiling:
## What

Staging builds resolve three targets at build time: the **engine gateway**, the **Agent Store gateway**, and the **client-metrics ingest**. CI only ever answered the first with a staging value. The other two fall back to a hardcoded production literal when unset:

- `ui/engine-client/src/store-catalog.ts:51` — `DEFAULT_STORE_GATEWAY`
- `app/src/lib/store-gateway-session.ts:19` — `STORE_GATEWAY_URL`
- `app/src/hooks/use-perf-spans.ts:14` — `CLIENT_METRICS_URL`

So the staging QA DMG **and** the preview web app browsed and **published agents into the real production catalog**, and folded QA timing data into the prod Prometheus histograms behind grafana.gethouston.ai, while everything else on those builds spoke to staging.

This bakes `VITE_AGENTSTORE_GATEWAY_URL` + `VITE_CLIENT_METRICS_GATEWAY_URL` from the same flavor-split expression the gateway URL already uses, in `build-macos` (release.yml), `build-web` (release.yml, prod), and `web-staging.yml` (preview).

The store var fixes both halves of the store leak on its own. `store-gateway-session` installs its `window.__HOUSTON_STORE__` override only when the hosted engine URL and the store URL **differ** — today they differ (staging engine vs prod store), which is exactly what redirected publishes to prod. Once they match, it steps aside, publish rides the engine base, and `store-catalog` reads the same var for browse.

## The empty-string trap this also closes

Both app-side call sites read their var with `??`, which only catches `null`/`undefined`. GitHub Actions **cannot omit a job-level `env:` key**, so `${{ cond && secret || '' }}` *sets* these to `""` on the legs that shouldn't carry them, and Vite's `loadEnv` copies empty `VITE_*` values through verbatim. Left as-is, adding these vars would have regressed the local `v*` channel: `STORE_GATEWAY_URL` collapses to `""`, the hook installs an empty base, and `storeApiBase()` falls through to the local sidecar, which serves no `/v1/agentstore` routes. Client metrics would post to a relative URL under `tauri://localhost`.

Both now route through a shared `bakedUrl()` that treats "set but blank" as absent.

## Prod is unchanged

On the prod legs these resolve to the same URL the hardcoded fallbacks already produced, and on the local `v*` channel they resolve to `""`, which `bakedUrl()` maps back to the same prod default. No behavior change on any shipping artifact.

## Verification

- Downloaded and mounted the shipped `Houston_staging_0.5.40_universal.dmg`; `strings` on `Contents/MacOS/houston-app` shows `staging-gateway.gethouston.ai` (the `option_env!` reads in the Tauri shell) and no prod gateway, confirming the engine URL was already correct and the store path simply had no staging value to read.
- `cd app && pnpm test` — 2670 pass, 0 fail (includes the new `baked-url.test.ts`).
- `pnpm typecheck` (all 27 workspace projects) and `pnpm check` clean.
- Both workflow files parse as YAML.

## Deliberately not fixed here

`VITE_AGENTSTORE_SITE_URL` (the `agents.gethouston.ai/a/<slug>` "view page" links). The staging store hostname lives in the private cloud repo and needs a new repo secret, and `ui/engine-client/src/client.ts:212` hardcodes it with no env escape hatch. Cosmetic, links only, nothing writes to the wrong place.

Also out of scope: `build-windows` / `build-linux` have no staging flavor at all, so there is no staging MSI or AppImage. Separate PR if we want them.

## How to confirm after the next cloud-v tag

Not by grepping the DMG. The frontend bundle is compressed inside `houston-app`, so these vars won't show up in `strings` (the gateway URL only did because Rust reads it too). Install the staging DMG, publish a throwaway agent, and confirm it lands in the staging catalog and **not** on agents.gethouston.ai.

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -31,7 +31,17 @@ VITE_CONTROL_PLANE_URL=http://localhost:9080
 #     dev app write to the REAL store under your own account. When the store
 #     has real traffic, flip back to the local loop:
 #     VITE_AGENTSTORE_GATEWAY_URL=http://localhost:9080 ------------------------
+#     This line is now the ONLY place the dev loop learns the store host: the
+#     client carries no hardcoded gateway fallback (a literal silently overrode
+#     whatever the environment meant, which is how the staging QA DMG published
+#     into the production catalog). Unset it and the dev app has no store.
 VITE_AGENTSTORE_GATEWAY_URL=https://gateway.gethouston.ai
+
+# --- Client perf spans (/v1/client-metrics): deliberately OFF in dev. Spans are
+#     still measured and mirrored to PostHog, just never shipped, so local runs
+#     don't skew the prod Prometheus histograms (they used to, via the same
+#     hardcoded fallback). Point it somewhere to exercise the ingest path:
+#     VITE_CLIENT_METRICS_GATEWAY_URL=http://localhost:9080 --------------------
 
 # --- Firebase (GCIP) project — public identifiers, not secrets. The one secret
 #     half (FIREBASE_API_KEY) lives in .env.local. -----------------------------
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -467,23 +467,25 @@ jobs:
       # channel gated: a non-cloud build has no gateway to authenticate against, so
       # it stays '' and lib.rs's empty-filter keeps integrations off there.
       HOUSTON_INTEGRATIONS_URL: ${{ startsWith(github.ref_name, 'cloud-') && (matrix.flavor == 'staging' && secrets.HOSTED_ENGINE_URL_STAGING || secrets.HOSTED_ENGINE_URL) || '' }}
-      # Agent Store gateway. Without this the store falls back to the hardcoded
-      # PRODUCTION default (`DEFAULT_STORE_GATEWAY` in
-      # ui/engine-client/src/store-catalog.ts, `STORE_GATEWAY_URL` in
-      # app/src/lib/store-gateway-session.ts), so the STAGING QA DMG browsed AND
-      # PUBLISHED against the real prod catalog while everything else on it spoke
-      # to staging. One var fixes both halves: store-gateway-session installs its
-      # `window.__HOUSTON_STORE__` override only when the hosted engine URL and
-      # the store URL DIFFER, so once they match on staging it steps aside and
-      # store-catalog reads this value for browse. On the prod leg this resolves
-      # to the same URL the hardcoded default already produced — no change to the
-      # shipping build.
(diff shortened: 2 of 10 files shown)


cravenceiling:
## Follow-up: the hardcoded gateway host is gone from the client entirely

First pass kept the pre-existing `"https://gateway.gethouston.ai"` literals and only added the missing bakes. That was wrong — the literal is the actual mechanism of this bug. A deployment hostname is configuration: it differs per environment and can move, and a literal doesn't "keep things working", it silently overrides whatever the environment meant. Staging had no store value to read, so the literal won and published into the production catalog.

All three gateway literals on the resolution path are deleted (`store-catalog.ts` `DEFAULT_STORE_GATEWAY`, `store-gateway-session.ts`, `use-perf-spans.ts`). The value now comes only from the build, and a build given nothing has no store, loudly:

- `storeCatalogApiBase()` throws a `StoreCatalogError` rather than guessing a host, so an unconfigured build shows an error in the store view instead of quietly browsing someone else's catalog.
- `store-gateway-session` installs no `window.__HOUSTON_STORE__`, so publish falls through to the engine base and fails visibly.
- `PerfSpanTransport.send` is now optional: no ingest URL means a mirror-only transport (spans still measured and mirrored to PostHog) and `flush()` drops the batch instead of queueing for a transport that will never exist.

`bakedUrl()` loses its fallback parameter and returns `undefined` for an absent or blank bake.

### Consequences worth reviewing

1. **CI must bake on every packaged leg now**, so `build-windows` and `build-linux` get the two vars as well. Both were relying on the literal without anyone noticing.
2. **The store bakes are NOT cloud-channel gated.** The Agent Store is one public catalog that local `v*` builds use too, so only the staging/prod flavor split applies. Previously the literal covered the local channel.
3. **A fork/CI build with no `HOSTED_ENGINE_URL` secret now has no store** instead of silently pointing at ours. Intentional, and `.env.development` documents the var for the dev loop.
4. **Client perf spans are now off in dev** (mirror-only). They were posting into the prod Prometheus histograms from every developer's machine via that same fallback.

### Still one literal left, and it needs a decision

`agents.gethouston.ai` — the store *site* host, for "view page" links — is hardcoded in three places (`app/src/components/store-view/use-my-store-agents.ts:31`, `ui/engine-client/src/client.ts:212`, `packages/web/src/engine-adapter/store-gateway.ts:38`). Killing it the same way needs a repo secret for the site host (there is no `AGENTSTORE_SITE_URL` secret today, unlike `HOSTED_ENGINE_URL`), so it needs someone to create that secret before the code change is safe. Links only, nothing writes to the wrong place. Say the word and I'll do it once the secret exists.

### Gates

- `cd app && pnpm test` — 2670 pass, 0 fail.
- `pnpm typecheck` (27 projects), `pnpm check`, `pnpm check:boundaries` all clean.
- `pnpm --filter @houston-ai/engine-client test` — 103 pass, 2 fail; both failures reproduce identically on `main` (`returns the page payload as-is`, `returns the creator page payload as-is` — a `deepStrictEqual` prototype quirk), so they predate this branch.

PR #1243, opened 2026-08-05
fix(providers): recognize NVIDIA's Public API Endpoints account gate (HOU-890)

cravenceiling:
## Summary

HOU-890: NVIDIA users fail with **410/404 status code (no body)** on chats and "We couldn't reach NVIDIA to check your key, so nothing was saved" at connect. Sentry: HOUSTON-APP-4WX (52 events / 22 users, still firing on 0.5.43).

### Root cause (verified live against a real key)

**NVIDIA serves each hosted model ("function") per account.** A live probe with a valid NGC Personal Key (Public API Endpoints scope enabled) showed the SAME key getting `404 {"detail":"Function '<uuid>': Not found for account '<id>'"}` on gemma / kimi / glm / nemotron while llama-3.3 / llama-3.1 / gpt-oss / minimax answered 200 — and `/v1/models` lists all 102 either way. Some accounts are missing the whole "Public API Endpoints" permission (every model gated; the NVIDIA forum wave since May), others just part of the catalog.

Houston's key verification probes ONE model — `modelFor()`, which for uncurated providers is the alphabetically-first catalog row: **`google/gemma-3-12b-it`, gated for many accounts**. So valid keys failed verification ("couldn't reach NVIDIA… try again", a retry loop that can't succeed), and the same model as chat default meant first chats died on the generic unknown-error card. The OpenAI SDK flattens NVIDIA's `{status,title,detail}` bodies (no `error` key) to `<status> status code (no body)`, which is why Sentry never showed the detail.

### Fix

- **Verify retries broadly-served models** (`packages/runtime/src/auth/nvidia-verify.ts`): a gated probe tries `meta/llama-3.3-70b-instruct` then `meta/llama-3.1-8b-instruct` before any verdict; a success verifies the key, a non-gate failure goes to normal classification, and only all-gated throws `key_restricted`. (`verify-api-key.ts` split with `verify-errors.ts` under the 200-line rule; stable exports re-served.)
- **NVIDIA default model** (`packages/runtime/src/ai/providers.ts`): `UNCURATED_DEFAULT_MODEL` pins nvidia to `meta/llama-3.3-70b-instruct` instead of the alphabetical first row, so new connects chat on a model that actually serves.
- **Mid-chat classification** (`packages/runtime/src/ai/provider-error.ts`): `isNvidiaFunctionGated` (nvidia + 404/410 + body-less / "function not found" / "not found for account") → `model_unavailable` with a one-click `meta/llama-3.3-70b-instruct` fallback — the switch-model card, since re-keying can never fix a gated model. A 404 with an unrelated body deliberately stays out.
- **Dialog copy** (en/es/pt): the residual all-models-gated error names the real remedy — check the key is an NGC Personal Key with "Public API Endpoints" under Services Included (org.ngc.nvidia.com), and NVIDIA support if it still fails. NVIDIA-specific override for `key_restricted` in `provider-api-key-dialog.tsx`.

### Before (reproduce)

1. Use an NVIDIA key whose account doesn't serve gemma (curl `/v1/chat/completions` with `google/gemma-3-12b-it` → `404 Function … Not found for account`, while `meta/llama-3.3-70b-instruct` → 200).
2. Settings → AI Models → NVIDIA → paste key → Connect → "We couldn't reach NVIDIA to check your key… try again in a moment", forever.
3. Mid-chat on a gated model: generic error card, `410/404 status code (no body)`.

### After (verify)

1. Same key connects: verification falls back to llama-3.3 and succeeds.
2. New NVIDIA connects default to `meta/llama-3.3-70b-instruct`; chats work.
3. Picking a gated model mid-chat renders the switch-model card with a one-click "Switch to meta/llama-3.3-70b-instruct" button.
4. A fully-gated account gets the honest key_restricted copy (NGC Personal Key check → NVIDIA support). Bad keys still read invalid_key (403 path pinned by tests).

### Tests

- `provider-error.test.ts`: 6 cases — 410/404 body-less and "Not found for account" (verbatim live shape, ids synthetic) → `model_unavailable` + fallback; fallback-on-fallback → null; unrelated-body 404 and other-provider 410 stay `unknown`.
- `verify-api-key.test.ts`: gated probe → fallback verifies (asserts the fallback model id); all-gated → `key_restricted`; gated-then-403 → `invalid_key`.
- Full runs: runtime 1011 ✓, host+domain 1271 ✓, app node:test 2700 ✓, tsgo app/web ✓, check-locales ✓, Biome ✓.

### DIFF
--- a/app/src/components/provider-browser/provider-grouping.ts
+++ b/app/src/components/provider-browser/provider-grouping.ts
@@ -201,6 +201,7 @@ export type ProviderDescriptionKey =
   | "google"
   | "amazon-bedrock"
   | "minimax"
+  | "nvidia"
   | "openai-compatible";
 
 /**
@@ -222,6 +223,7 @@ const DESCRIPTION_KEY_BY_ID = {
   google: "google",
   "amazon-bedrock": "amazon-bedrock",
   minimax: "minimax",
+  nvidia: "nvidia",
   "openai-compatible": "openai-compatible",
 } satisfies Record<string, ProviderDescriptionKey>;
 
--- a/app/src/components/shell/provider-api-key-dialog.tsx
+++ b/app/src/components/shell/provider-api-key-dialog.tsx
@@ -56,6 +56,17 @@ const REASON_COPY: Record<
   provider_unavailable: "apiKey.errorProviderUnavailable",
 };
 
+/**
+ * NVIDIA's `key_restricted` is an ACCOUNT gate, not key settings: NVIDIA has
+ * to enable "Public API Endpoints" on the account's org, so the generic
+ * "create a new key" remedy would send users in circles (HOU-890).
+ */
+function reasonCopyKey(providerId: string, reason: ApiKeyConnectReason) {
+  if (providerId === "nvidia" && reason === "key_restricted")
+    return "apiKey.errorNvidiaAccountGated" as const;
+  return REASON_COPY[reason];
+}
+
 export function ProviderApiKeyDialog({ provider, onClose }: Props) {
   const { t } = useTranslation("providers");
   const [key, setKey] = useState("");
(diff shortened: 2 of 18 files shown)


cravenceiling:
Third commit (088605a6): guided key-creation UX. The dialog's "Get your API key" button now opens the NGC Personal Key page (org.ngc.nvidia.com/setup/api-keys) instead of build.nvidia.com's quick flow (which never offers the required scope), and a per-provider guide panel walks the user through the three steps: sign in, Generate Personal Key with "Public API Endpoints" under Services Included, paste the nvapi- key. en/es/pt. Also fixes the raw "providers.nvidia.description" i18n key in the provider modal. New e2e spec (nvidia-connect-guide.spec.ts) guards the guide; screenshots verified in both themes; visual baselines unchanged (8/8 pass).

PR #1250, opened 2026-08-06
fix(app): stop Windows shortcut icons going blank after MSI updates (PRODUCT-1233)

cravenceiling:
Fixes PRODUCT-1233 (app icon disappearing on Windows — taskbar pins and desktop).

## Root cause

Houston ships Windows as a WiX MSI and auto-updates in the background. Two defects in the stock Tauri bundler setup:

1. **Taskbar pins (deterministic).** The stock Start Menu shortcut sets `Icon="ProductIcon"`. An MSI shortcut icon is extracted to `C:\Windows\Installer\{ProductCode}\` and referenced by absolute path from the `.lnk`. Every Houston release is a WiX MajorUpgrade with a fresh `ProductCode` (`Product Id="*"`), and Windows deletes the old folder on upgrade — so any pin the user made from that shortcut keeps pointing at a deleted icon file forever. First background update after pinning ⇒ blank white icon, which is why users see it happen "out of nowhere".
2. **Desktop icons (racy).** `MajorUpgrade Schedule="afterInstallInitialize"` uninstalls the old product (removing `Houston.exe`) before laying down the new one. If Explorer re-resolves shortcut icons during that window it caches the blank fallback for the exe path, and the stale cache entry sticks.

Upstream Tauri still has the `Icon="ProductIcon"` attribute on `dev`, so no CLI bump fixes this.

## Fix

- **`app/src-tauri/wix/main.wxs`** — vendored copy of the tauri-cli v2.11.4 MSI template (wired via `bundle.windows.wix.template`) with one deliberate change: the Start Menu shortcut no longer sets `Icon="ProductIcon"`, so shortcuts (and pins made from them) take the icon embedded in `Houston.exe`, whose path is stable across updates. `ARPPRODUCTICON` (Apps-list icon) is untouched. Provenance + re-sync instructions are in the file header. *(Vendored upstream template, hence >200 lines — one attribute changed.)*
- **`app/src-tauri/src/windows_icon_repair.rs`** — heals already-broken installs. Once per installed version (first launch after every update, exactly when breakage can occur), a background thread rewrites the icon location of any user `.lnk` (taskbar-pin folder + desktop) that targets the current exe but references a missing icon file or one under `C:\Windows\Installer\`, then runs `ie4uinit.exe -show` to refresh Explorer's icon cache (the same post-update refresh Squirrel/Electron use). PowerShell runs via `-EncodedCommand` (no argv re-quoting), hidden window, install dir as cwd.

## Verification

**Repro before the fix** (Windows):
1. Install Houston from the MSI, pin it to the taskbar from the Start Menu entry.
2. `Get-ItemProperty` isn't needed — inspect the pin: `(New-Object -ComObject WScript.Shell).CreateShortcut("$env:APPDATA\Microsoft\Internet Explorer\Quick Launch\User Pinned\TaskBar\Houston.lnk").IconLocation` → points into `C:\Windows\Installer\{GUID}\`.
3. Install the next version's MSI (or let the auto-updater run). The referenced folder is gone; the pinned icon renders as the blank default.

**After the fix:**
1. Build an MSI from this branch (`cd app && pnpm tauri build --target x86_64-pc-windows-msvc`), install, and check the new Start Menu shortcut: its `IconLocation` is empty/exe-based, not `C:\Windows\Installer\...`.
2. On an install with an already-blank pin: launch the app once, wait a few seconds — the pin's `IconLocation` is rewritten to `<path>\Houston.exe,0` and the icon renders again (`[icon-repair]` lines appear in the app log).
3. Upgrade again from a pinned state: icon survives.

## Tests

- 4 new unit tests for the repair gating + PowerShell script construction (`cargo test windows_icon_repair`; full shell suite: 216 passed).
- `cargo check` clean on macOS and `--target x86_64-pc-windows-gnu` (compiles the `cfg(windows)` path).
- Biome clean on the config edit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2075,6 +2075,7 @@ name = "houston-app"
 version = "0.5.41"
 dependencies = [
  "arboard",
+ "base64 0.22.1",
  "bytes",
  "chrono",
  "dirs 5.0.1",
--- a/app/src-tauri/Cargo.toml
+++ b/app/src-tauri/Cargo.toml
@@ -114,6 +114,9 @@ notify-rust = "4.11"
 # file is useless to another user. See `src/auth.rs` storage module.
 [target."cfg(target_os = \"windows\")".dependencies]
 windows-sys = { version = "0.59", features = ["Win32_Foundation", "Win32_Security_Cryptography", "Win32_System_JobObjects", "Win32_System_Threading", "Win32_UI_Shell"] }
+# PowerShell -EncodedCommand payload for the icon-repair pass
+# (src/windows_icon_repair.rs) — dodges Windows argv re-quoting.
+base64 = "0.22"
 # Same reason as notify-rust on Linux, but notify-rust's action handling is
 # XDG/D-Bus-only — on Windows we use the WinRT toast's `on_activated` callback
 # to catch the click. This is the crate notify-rust uses under the hood.
--- a/app/src-tauri/src/lib.rs
+++ b/app/src-tauri/src/lib.rs
@@ -19,6 +19,7 @@ mod oauth_loopback;
 mod shell_env;
 mod store_deep_link;
 mod window_focus;
+mod windows_icon_repair;
 
 use engine_supervisor::{
     reserve_free_port, resolve_engine_binary, spawn_supervisor, wait_until_host_healthy,
@@ -323,6 +324,12 @@ pub fn run() {
                     );
                 }
             }
+
+            // MSI upgrades break taskbar-pin/desktop icons (PRODUCT-1233);
+            // once per installed version, repair stale shortcut icons and
+            // refresh the shell icon cache on a background thread.
+            #[cfg(target_os = "windows")]
+            windows_icon_repair::spawn_repair(app.package_info().version.to_string());
             {
                 use tauri_plugin_deep_link::DeepLinkExt;
                 let handle = app.handle().clone();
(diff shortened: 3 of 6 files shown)


cravenceiling:
Ran an independent second-opinion review (Codex, adversarial) of the diagnosis and fix. Verdict: the pinned-shortcut root cause and the template fix are confirmed correct (non-advertised shortcut without `Icon` falls back to the target exe's icon per the Shell Links contract; `ARPPRODUCTICON` unaffected). Hardening it suggested is now in `9c96a33`:

- sweep the **Public Desktop** too — the per-machine MSI writes its desktop shortcut to `%PUBLIC%\Desktop`, not the user's
- parse `IconLocation` from the **last** comma, trim quotes, and expand env vars before existence checks

Two review notes that need the Windows VM loop (knowledge-base/windows-testing.md) before merge rather than more code:

1. **Verify `WScript.Shell` `.Save()` preserves `System.AppUserModel.ID`** on a rewritten pin (grouping identity). If it doesn't survive, the repair should move to native `IShellLinkW` + `IPropertyStore`.
2. The desktop-blanking leg is a timing hypothesis (exe absent during the MajorUpgrade's uninstall window); the repair + `ie4uinit -show` covers it either way, but a pre/post-upgrade dump of `IconLocation` on an affected machine would settle it.

Also confirmed: `icon.ico` (16/24/32/48/64/256 PNG entries) is adequate — a bad ico is ruled out as the cause.

PR #1258, opened 2026-08-06
fix(chat): markdown file links open and preview as rendered documents (PRODUCT-1231)

cravenceiling:
Fixes PRODUCT-1231.

## Root causes

Opening a markdown deliverable from chat was broken in three separate places, all of which a user experiences as "markdown preview is not working".

**1. `rehype-harden` blocked the most common link shape.** It treats a destination as relative only when it starts with `/`, `./` or `../`. A bare `[Perfil](perfil.md)` — exactly what agents write right after writing a file — parsed as neither absolute nor relative, so the `<a>` was replaced by an inert grey `<span>` with `" [blocked]"` appended. The user's own file was named in chat as dead text, and the `a` override never even ran.

**2. Percent-encoding was never undone.** micromark normalizes every link destination through `normalizeUri`, so a name with a space arrived as `Tropical%20Food%20-%20Plan.md`; harden then rewrote surviving hrefs again (`./a.md` → `/a.md`, re-encoded). That names no file the host has, so the preview dialog opened straight into "Couldn't load the file" and titled itself with the escaped text — the screenshot on the issue.

**3. Markdown previewed as raw source.** A `.md` opened from chat is the deliverable the agent just wrote, but it rendered as monospace source in a `<pre>`.

**4. Content overflowed the modal.** `DialogContent` was a bare `grid`. An implicit auto track refuses to shrink below its content's min-content width, so one nowrap child — the `truncate` title, an unbreakable word, a wide table — pushed the whole surface past its max-width.

## The fix

- **`ui/chat/src/file-link-rehype.ts`** (new) runs between `sanitize` and `harden`. Order is load-bearing: after `sanitize`, whose `a` attribute whitelist would strip what it mints, and before `harden`, so it can hand harden a shape harden accepts. It records the pristine decoded path on `data-file-path` and rewrites a bare destination to `./`-prefixed form. Harden still owns and still rewrites the `href`; nothing here weakens it.
- **The `a` override** now opens `data-file-path` rather than harden's mangled href.
- **`app/src/components/file-preview-body.tsx`** (new, extracted so the dialog stays a thin shell) renders `.md` as formatted prose through the same Streamdown pipeline chat uses, with links inline rather than as chat's labeled button pills — a filled black slab mid-paragraph is the wrong grammar for a document.
- **`ui/core` dialog** uses `grid-cols-[minmax(0,1fr)]`; the preview frame clips, and code blocks and tables scroll inside their own frames.
- **`decodeMarkdownHref`** in `app/src/lib/agent-file-paths.ts` guards the app-side href seam too. Both decoders leave malformed escapes alone, so a file genuinely named `100%.md` still resolves.

## Verification

**Before the fix**
1. `pnpm dev`, open an agent chat.
2. Have the agent (or seed via `/__test__/chat-history`) emit `Documento listo: [Estrategia](<Tropical Food - Estrategia Integral 90 Dias.md>)` alongside a bare `[el plan](perfil.md)`.
3. The bare link renders as grey `el plan [blocked]` and is not clickable.
4. Clicking the pill opens a dialog titled `Tropical%20Food%20-%20Estrategia%20Integral%2090%20Dias.md`, whose title overruns the modal's right edge, showing "Couldn't load the file".

**After**
3. Both links render as working pills.
4. The dialog is titled `Tropical Food - Estrategia Integral 90 Dias.md` (truncated with an ellipsis, clear of the close button) and renders the file as formatted markdown — headings, bold, tables, lists, code blocks. Nothing paints outside the modal.

Automated:
- `cd app && pnpm test` — 2731 pass
- `pnpm --filter @houston-ai/chat test` — 370 pass (9 new, covering the rehype pass)
- `pnpm --filter houston-web test:e2e` — 253 pass, incl. the new `chat-markdown-preview.spec.ts`, whose second test walks every element in the dialog and asserts none extends past its box
- `pnpm --filter houston-web test:visual` — 8 pass, baselines unmoved despite the shared dialog change
- `pnpm check`, `pnpm typecheck`, `pnpm check:boundaries`, `pnpm check:parity`, `pnpm check-locales` all clean

## Known gap (not fixed here)

`[label](Tropical Food - Plan.md)` with **unescaped spaces and no angle brackets** is not a link per CommonMark and renders as literal text. Repairing that means rewriting the markdown source before parsing, which risks false positives inside code fences — worth its own issue, and arguably better addressed in the agent prompt (`<...>` or `%20`).


cravenceiling:
## Follow-up in this PR: a file is no longer dressed as a web link

The bug fix made file links *work*. Reviewing the result surfaced a design problem underneath: chat rendered the same action two different ways, and both borrowed the **web** vocabulary.

- `[Perfil](perfil.md)` → filled black button pill + `↗`
- `[plan.md](plan.md)` → blue autolink chip + 🔗

One action, two affordances — and `↗` / link-blue both mean "this leaves the app," which is the opposite of what a workspace file does.

### The rule

The affordance now follows the **destination kind**, before anything about the label. `data-file-path` (already minted by the file-link rehype pass) is exactly the signal "this destination is a file," so every file takes one branch and renders one chip: the per-extension glyph in its reserved `filetype.*` tint on the recessed chip surface — the same mark the Files tab and the turn summary already use. Label shape still decides among URLs, where it genuinely matters.

Label is the agent's own when it wrote one, the file name when it wrote the path as the label; the full path is always the chip's `title`.

### Why it matters beyond consistency

`DESIGN.md` reserves `text-link` blue as "the ONE sanctioned blue." Keeping it strictly for the web means **blue now reads as "leaves Houston" and a neutral chip as "this is yours"** — a distinction the reader could not make at all before.

This composes well with **PRODUCT-1152 (#1259)**, which landed on main mid-review and unified every URL onto the Slack chip. Merge conflict in `message.tsx` resolved by keeping both: their single treatment for every URL, with the file branch ahead of it. Their retirement of the button pill actually sharpens the contrast.

### Structure

`file-type.ts` + `file-type-icons.tsx` moved from `ui/agent` to `@houston-ai/core` — both the Files surfaces and chat prose now draw files, and the two must not drift. `FileTypeGlyphInline` is new: text-sized and **tinted**, unlike the monochrome card-body glyph, because inline there is no tile above it carrying the type.

Inventory **v54** (`assistant-message` anatomy gains `file-chip`) + changelog entry.

### Verification

CI green on `edd483d5` (run 31135464350) — all 12 jobs. Locally: app 2766, ui/chat 372, web e2e 257, visual 8 (baselines unmoved), plus check / typecheck / boundaries / parity / locales. New e2e asserts a labeled and a bare file link both render as one chip carrying the full path, that the chip is neither an anchor nor on the link tint, and that the web link beside it still is.

PR #1274, opened 2026-08-07
fix(integrations): progressive app-scoped discovery for integration search (PRODUCT-1274)

cravenceiling:
Fixes PRODUCT-1274.

## Root cause

`integration_search` had one shape: a single free-text query merged from a connected-toolkits lookup, a global lookup, and catalog name resolution. When a task named an app ("in PostHog, get the top 100 users by activity"), Composio's full-text ranking let unrelated connected apps (GitHub) dominate both lookups; the named app surfaced only as a toolkit row with no actions, and the agent concluded no tools existed — even though a later exact search proved they did. There was no way to scope discovery to the named app and no deterministic fallback when full-text scored the phrasing at zero.

## Fix (progressive discovery)

- **Explicit hard scope**: `integration_search` gains an optional `app` param ("posthog", "google sheets"…). It threads runtime tool → `/sandbox/integrations/search` → `IntegrationProvider.search(userId, query, acting, app)` → every provider. The Composio adapter resolves it against the catalog (`resolveScopeToolkits`: exact normalized match wins, else both-way substring) and returns ONLY that app's actions.
- **Deterministic fallback**: a scoped query that scores zero degrades to listing the toolkit's actions directly (`/tools?toolkit_slug=…`), so a named app always surfaces its real slugs — never an empty result from weak full-text. An unresolvable scope (typo/guess) returns empty from every provider; the sandbox proxy alone retries the fan-out unscoped once and flags the response `unscopedFallback: true`, which the tool renders as a leading "no app matching X — these are OTHER apps" note (never a false "no such app", and never provider-internal fallback noise polluting the merge).
- **Auto-scoping**: even without `app`, an app named in the query text gets its own toolkit-scoped lookup ranked first, ahead of connected-scoped and global noise (no listing fallback there, so a loose text match can't flood the result).
- **Custom provider parity**: `searchCustomTools` honors the same scope (integration filter + listing fallback + app row for a toolless integration).
- **Gateway adapter** (`remote.ts`) forwards `app` verbatim; a gateway that predates the field ignores it and serves today's unscoped result. The managed cloud's Go gateway needs the matching scoping as a follow-up to get the hard filter there.
- **Agent policy**: the tool description + empty-result messages now teach the retry ladder — split independent needs into separate searches, always pass `app` when the user names one, re-search scoped when an app shows only an app row, and never report a capability gap from a noisy cross-app result.

Module split (200-line rule): pure name/scope resolution moved to `composio-resolve.ts`, status stamping to `composio-annotate.ts`; `composio-search.ts` keeps only the discovery policy.

## Reproduce (before)

1. Connect both GitHub and PostHog to an agent (direct-adapter deployment: `pnpm dev` with `COMPOSIO_API_KEY` set).
2. Ask the agent: "In PostHog, get the top 100 users by activity."
3. `integration_search("in posthog get the top users by activity")` returns GitHub actions dominating; PostHog appears only as an app row with no actions, and the agent reports it can't do it.

## Verify (after)

1. Same setup; same request. The search result now leads with PostHog actions (the query-scoped lookup ranks first).
2. Scoped search: call the tool with `app: "posthog"` (or watch the agent do it) — the result contains ONLY PostHog actions; when full-text scores zero, the direct toolkit listing still returns real slugs.
3. Scoped search for a nonexistent app returns the not-found message with a spelling-retry instruction, not a policy claim.
4. `pnpm --filter @houston/host --filter @houston/runtime test` — new coverage in `composio-search.test.ts`, `custom/search.test.ts`, `remote.test.ts`, `routes/integrations.test.ts`, and the runtime tool tests, all green (host 1348 / runtime 1026). `pnpm check`, `pnpm typecheck`, `pnpm check:boundaries` all clean.

Out of scope, noted for follow-up: Composio's session-reuse discovery API (issue point 4) and the closed Go gateway's matching `app` scoping.


cravenceiling:
Independent Codex review (gpt-5.6-sol, high effort) of the initial diff returned FIX-FIRST with 4 findings, all real and all addressed in 3fe6d108:

1. **P1 — per-provider unscoped fallback polluted the fan-out merge**: a scope Composio couldn't resolve triggered its internal unscoped search, ranking noise ahead of the custom provider's correctly scoped hits (and a custom-only install got a false not-found instead of any fallback). Providers now return empty on an unresolved scope; the sandbox proxy alone retries the fan-out unscoped once when ALL providers merged empty, marking the response `unscopedFallback: true`, which `integration_search` renders as a leading "no app matching X — these are OTHER apps" note.
2. **P1 — loose scope could hard-scope to several unrelated apps** ("hub" → github + hubspot, up to 50 listed actions each): non-exact resolution now yields only the single closest candidate.
3. **P1 — custom provider lacked exact-match precedence** ("Acme" also pulled "Acme Staging"): exact normalized slug/name match now wins outright.
4. **P2 — whitespace-only `app` became a real scope** (false not-found on custom-only hosts): trimmed to undefined at both routes.

Each fix is pinned by a test; host 1351 / runtime 1027 green, Biome + typecheck + boundaries clean. One doc correction from the earlier orphaned review pass also landed in 45109d6f (the port contract note; superseded by the strict-scope wording in 3fe6d108).

cravenceiling:
Pushed 511383da addressing the review findings (and the live dev-loop repro where searches through the Go gateway still dumped every connected app and burned 3 scoped re-searches):

- **Scope-status handshake**: `IntegrationProvider.search` → `{ items, scope: resolved|unresolved|ignored }`; the user route echoes `scoped: true|false`; the remote adapter reports **ignored** when the upstream echoes nothing (today's gateway). Sandbox proxy: all-unresolved → the one unscoped retry (`unscopedFallback`), any-ignored → `scopeIgnored` and no retry. The runtime tool never claims "no such app exists" (or hard scoping) over results a pre-contract gateway never scoped.
- **One shared resolver** (`scope-resolve.ts`) for Composio/custom/fake: exact match now beats the <3-char guard ("HR" resolves), slug-aware near tie-break, custom capped to single closest loose match.
- **Token-run query resolution**: "inbox" no longer resolves the app "box"; unconnected query-resolved apps rank after connected hits.
- **Connected named app + zero-scoring phrasing lists its actions in ONE search** (the "En Clockify…" case) — no more model-driven re-search ladder.
- Dedup cleanups: `annotate()` for toolkit rows, shared `optionalTrimmed`, dead `limit` param removed.

Host 1364 / runtime 1029 / domain 191 tests green; biome + boundaries clean. The Go gateway's matching `app` scoping + `scoped` echo is in flight in the cloud repo on branch `product-1274-gateway-app-scope`.

PR #1289, opened 2026-08-10
fix(runtime): serve-sync every pi api-key provider so uncurated keys survive pod recycles (PRODUCT-1213)

cravenceiling:
## Root cause

PRODUCT-1213: Cerebras (and every other **uncurated** pi api-key provider) reads disconnected in the cloud app after a pod recycle, even though the user never logged out.

- Connect was deliberately widened to accept a pasted key for **any** pi api-key provider (`packages/host/src/providers/api-key.ts`), and the gateway durably stores it (verified in prod: the `gateway.org_credentials` row and the GSM secret for Daniel's cerebras key are both intact and enabled).
- But `runServedSync` (`packages/runtime/src/auth/serve.ts`) probed only the **curated** `PROVIDERS` catalog (11 ids). An uncurated provider's key therefore lived only in the pod-local `auth.json` written at connect time — a file deliberately excluded from store-sync.
- On the next pod roll the runtime rebuilt `auth.json` exclusively from the serve sync, which never asked for cerebras → the provider vanished from `/providers` and the app showed it disconnected. Verified live on prod pod `agent-09638b2bed09ece1`: `auth.json` / `served-providers.json` held only `anthropic, openai-codex, openrouter` while the gateway still held the cerebras key.

## Fix

Probe the curated catalog **plus every uncurated pi api-key provider id** (`piApiKeyProviderIds()`, the runtime mirror of the host's `isApiKeyProvider` connect gate): whatever connect can store centrally, the serve sync can now re-hydrate. The host's `/sandbox/credential` route and the gateway's pod-credential route were already provider-generic — no changes needed there. Desktop/self-host is untouched (`serveModeOn()` is false).

## Before (repro on prod ≥ v0.5.43)

1. In the cloud app, connect Cerebras with an API key on the AI Models page; chat with a Cerebras model — works.
2. Wait for the agent pod to be recycled (engine roll / pod sleep), or delete the pod.
3. Reopen the app → Cerebras shows disconnected; the gateway still holds the key (row in `gateway.org_credentials`, enabled GSM secret version).

## After

1. Same connect + chat.
2. Recycle the pod (or wait for the next engine roll).
3. Reopen the app → Cerebras still shows connected and chats work; pod `auth.json` re-hydrates the key on the first `/providers` poll (`[serve] applied central credentials: ... cerebras`).

## Tests

- New: `serve.test.ts` — an uncurated provider (cerebras) served centrally is probed, applied to `auth.json` as an `api_key` entry, and manifest-tracked.
- Updated: the shared-in-flight sync test's probe count now covers curated + uncurated ids.
- `pnpm --filter @houston/runtime test` (1044 pass), `pnpm check` (Biome clean), `pnpm typecheck` (all 27 projects pass).


cravenceiling:
CI note: the shard-5 e2e failure was **not** caused by this PR — main has been red since a17caae2 (#1286). `fd8204bf` renamed the create dialog's from-scratch card from "Create new agent" to "Create new" and updated `support/sidebar-create.ts`, but not `support/create-agent.ts` / `agent-onboarding.spec.ts`'s local copy, so every spec walking the create-agent flow timed out (sidebar-dnd on shard 5; agents/agent-onboarding elsewhere — same failure on main runs 31412414149 and the in-flight 31422641488).

Fixed here in 8f4e4466 by matching the tile via role+exact name. Full e2e suite green locally, 297/297.

PR #1297, opened 2026-08-11
fix(credentials): stop revoked-credential resurrection loops + Sentry noise (HOUSTON-APP-530)

cravenceiling:
Fixes Sentry [HOUSTON-APP-530](https://houston-cd.sentry.io/issues/7638934035/) — `[sandbox/credential] revoked-token report for anthropic: credential disconnected` — 734 events / 380 users since 2026-07-28, still occurring.

## Root cause

Two defects stacked, neither of which is a mass credential-revocation bug at Anthropic:

**1. The success path logged at error level.** The Sentry event IS the `console.error` in `routes/credential-revoked.ts` for the `removed:true` branch — the branch whose own comment says "both outcomes are success". The host's Sentry console capture turns every `console.error` into an event (`knowledge-base/sentry.md` explicitly bans error-level operational lines), so every *working, expected* provider-revocation disconnect (HOU-952 doing its job: delete the dead credential, show the reconnect card) paged as an error. The issue's firstSeen (2026-07-28) is the day after the report route shipped (d899178f, 2026-07-27) — the volume began when the logging shipped, not when revocations did.

**2. Revoked credentials get resurrected, looping the delete.** Event analysis shows heavy repeaters: one user with 67 confirmed deletes, bursts of 4 deletes 15–90 seconds apart from a single pod (e.g. 23:37:59 → 23:38:33 → 23:38:49 → 23:39:07). Every `removed:true` requires the store to hold the dead token again — something keeps refilling it. The refill channels:

- **Old desktop clients' cached-snapshot reconcile** (`?if_absent=1` claude-oauth push). The client side was deleted in HOU-954/HOU-950 (8bb2b88b, 2026-07-27) but "the host keeps honoring if_absent for older clients" — and after a revoked-report delete the row is absent, so the fill is accepted. A cached snapshot of the revoked family poisons it again (Anthropic's refresh-token-reuse detection, the HOU-855 mechanism), the next turn 401s "has been revoked", reports, deletes, and invites the next fill.
- **Pod-side legacy-fallback adoption** (`RemoteCredentialStore.adoptFallback`): any team-scope serve miss re-uploads the pod-local legacy credential file with an insert-only PUT — including a miss caused by a revocation delete.
- **The serve healer** (`CredentialServeHealer`): re-uploads a refresh-bearing runtime credential on a serve miss, revocation-blind.
- **Sub-minute bursts from one pod** additionally implicate control-plane DELETE→GET lag re-serving a just-deleted row (gateway-side, closed repo — see "What stays server-side" below). Each re-serve burned a user turn and re-fired the report.

## Fix

**`credentials/revocation-tombstones.ts` (new):** a confirmed revocation delete tombstones that (workspace, scope, provider) for 15 minutes, and every *automatic* refill path is refused while it is live:

- the `ifAbsent` claude-oauth fill throws `RevokedRefillBlockedError` → routes answer **409** (deliberate refusal, not a broken hop),
- `adoptFallback` skips adoption,
- the healer refuses the attempt,
- `disconnectRejectedCredential` (the invalid_grant twin of the revoked report — same terminal death, same refill loop, HOU-855's exact cycle) also marks the tombstone.

User-driven connects are never blocked and **clear** the tombstone: an overwrite claude-oauth push (fresh browser login), a verified api-key/setup-token paste, a device-code capture. The tombstone is deliberately scope-blunt rather than digest-scoped: a cached snapshot of the revoked family carries a *different* access token than the digest the report named (the gateway rotated the family since it was cached), so a digest tombstone would block nothing.

**Logging split by what it means:**
- confirmed removal → `console.info` (pipeline working; the user gets the reconnect card — no Sentry event),
- superseded report → `console.info` (unchanged semantics),
- a **second** confirmed removal inside the tombstone window → `console.error` naming the resurrection — the one outcome that is a real defect, now a precise, low-volume Sentry signal instead of 734 noise events.

**Runtime-side burst damper** (`auth/report-revoked.ts`): one report per (scope, provider, token-digest) per pod lifetime, mirroring `served-key-guard.ts` — a failed report un-marks so a later turn retries. Stops one pod amplifying a lagging control plane into a burst of confirmed deletes.

## What stays server-side (documented, not fixed here)

The sub-minute same-pod delete bursts cannot be fully explained by open-repo channels (the old client reconcile is once-per-app-session; heal was observed failing between burst events). The remaining suspect is the gateway's own DELETE→GET consistency (replica/cache lag re-serving a just-deleted row, or its `removed:true` answer for an already-gone row). The new escalation log line ("refilled and revoked AGAIN within the tombstone window") plus the runtime dedupe will make any residual gateway-side resurrection show up as a small, precisely-labeled Sentry issue to hand to the gateway team.

## Tests

- `credentials/revocation-tombstones.test.ts` (new): TTL, mark-returns-active resurrection signal, personal/team key semantics, clear.
- `routes/credential-revoked.test.ts`: removal marks tombstone + logs info not error; second removal escalates; superseded leaves no tombstone.
- `routes/claude-oauth.test.ts`: tombstoned `if_absent` fill → 409, nothing written; fresh overwrite clears and fills resume.
- `credentials/remote-store.test.ts`: no fallback adoption while tombstoned (GET only, no PUT).
- `routes/credential-healer.test.ts`: heal refused while tombstoned, resumes after clear without spending the cooldown.
- `credentials/disconnect.test.ts`: dropped rejection marks tombstone; superseded does not.
- `runtime auth/report-revoked.test.ts`: same digest reported once per pod lifetime; new digest reports; failed report retries.

`pnpm --filter @houston/host --filter @houston/runtime --filter @houston/domain test`, `pnpm check` (Biome), `pnpm check:boundaries`, and both packages' `typecheck` are clean. `knowledge-base/anthropic-credentials.md` documents the tombstone under the disconnect policy section.

### DIFF
--- a/knowledge-base/anthropic-credentials.md
+++ b/knowledge-base/anthropic-credentials.md
@@ -170,6 +170,34 @@ way the coalescer's cached result for that key is forgotten, and a failure of th
 re-read itself degrades to "absent" rather than escaping as a 500 in place of the
 404 the runtime knows how to act on.
 
+### A provider-revoked credential must STAY deleted — the revocation tombstone
+
+The revoked-token report deleting the central row is only half the job; Sentry
+HOUSTON-APP-530 (734 events / 380 users in two weeks) was the other half:
+everything that put the dead credential BACK. Old pre-HOU-950 desktop builds
+still loop the cached-snapshot reconcile (`?if_absent=1` claude-oauth push) on a
+15–30s cadence, and the pod itself could refill via the legacy-fallback adoption
+(`credentials/remote-store.ts`) or the serve healer. Every refill of a revoked
+family fails the next turn, which reports, which deletes, which invites the next
+refill — one turn burned and one Sentry error per cycle, forever.
+
+`credentials/revocation-tombstones.ts` breaks the cycle: a CONFIRMED removal
+(`removed:true` on `/sandbox/credential/revoked`) tombstones that
+(workspace, scope, provider) for 15 minutes, and every AUTOMATIC refill path
+consults it — the `ifAbsent` claude-oauth fill answers 409, fallback adoption
+and the healer skip. Deliberately scope-blunt, not digest-scoped: a cached
+snapshot of the revoked family carries a DIFFERENT access token than the one
+the report named (the gateway rotated the family since), so a digest tombstone
+would block nothing. USER-driven connects are never blocked and CLEAR the
+tombstone (overwrite claude-oauth push, verified api-key/setup-token paste,
+device-code capture); the TTL bounds any clear-site we missed. Logging split
+accordingly: a confirmed removal is the pipeline WORKING (the user gets the
+reconnect card) and logs info; the error-level line — the only one Sentry
+should page on — is a SECOND confirmed removal inside the tombstone window,
+which means something (a stale client, a path that bypasses the pod host)
+resurrected a dead credential past the guards. The ledger is per-pod-process
+and in-memory: a pod restart forgives it, costing at most one extra cycle.
+
 ## The six traps (each was a live bug, or a guarded near-miss)
 
 1. **`USER` must reach the SDK subprocess.** The CLI names its Keychain
--- a/packages/host/src/channel/proxy.ts
+++ b/packages/host/src/channel/proxy.ts
@@ -1,6 +1,11 @@
 import type { IncomingMessage, ServerResponse } from "node:http";
 import type { ClaudeOAuthCredential, CustomEndpoint } from "@houston/protocol";
 import {
+  type RevocationTombstones,
+  RevokedRefillBlockedError,
+  sharedRevocationTombstones,
+} from "../credentials/revocation-tombstones";
+import {
   ApiKeyRejectedError,
   type CaptureResult,
   type ChannelCtx,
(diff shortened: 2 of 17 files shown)


cravenceiling:
Reviewed against tonight's HOUSTON-APP-56F root-cause work (#1293) and resolved the conflicts.

**Relationship to #1293 — complementary, not the same fix.** One correction to this PR's evidence: the "sub-minute same-pod delete bursts" (the 23:37:59→23:39:07 example) were NOT gateway DELETE→GET lag — that was the #1293 bug live (the acting-scope 502 sending a user around the reconnect carousel; each fresh sign-in overwrote the row and its family died to serial-mint collateral). With #1293 merged, that driver is gone, so expect this issue's volume to collapse further than projected. The tombstone remains right for the channels that do still refill: pre-HOU-950 clients' `if_absent` pushes, fallback adoption, and the healer.

**Review findings (verified, no changes needed):**
- Tombstone keys derive from the acting token's `sub` (`credentialScopeKey`), so two different minted tokens for the same user hit the same key — personal tombstones actually gate. Verified against `scope-key.ts`.
- `clear()` sites cover every user-driven connect path (overwrite push, verified api-key/setup-token, device-code capture), so a real reconnect is never blocked and can't strand a tombstone.
- The digest-guarded compare-and-delete semantics are untouched; the escalation only fires on a genuine automatic refill (a user reconnect clears first, so a fresh credential dying inside the window logs info, not the resurrection error).

**Merge resolution (10c03e3d):** #1293's attributed-connect early-return skips the materialize hop, so the overwrite tombstone `clear()` now runs BEFORE that return — every cloud connect is attributed, and clearing after would have stranded live tombstones on every cloud reconnect (auto-recovery blocked for a TTL + a primed false escalation). Added a test pinning the ordering: an attributed overwrite clears personal+team tombstones (two different tokens, same sub) while still never touching the runtime.

Gates: host 1428 / runtime 1068 / domain 193 tests green, Biome clean, typecheck clean, boundaries clean.

PR #1298, opened 2026-08-11
fix(credentials): stop the dead google key serve/adopt loop (HOUSTON-APP-567)

cravenceiling:
Fixes HOUSTON-APP-567 — `[serve] served google credential is not an API key (an OAuth-type token that can never authenticate)`, 409 events / 327 users on managed-cloud since 2026-08-04, still escalating.

## Root cause

The events began the day the HOU-1107 serve guard (#1198) first deployed (engine-pod release `67cc0166`, 2026-08-04) — the guard doing its job over legacy pre-verification rows where users pasted Google OAuth tokens (`ya29.…`, JWTs) as Gemini API keys. It should have been a one-shot healing wave, but the event distribution shows agents firing repeatedly for days (one fired 25 times; one pair of events landed 2 minutes apart). Two loops kept the dead rows alive:

1. **Resurrection through 404-adoption.** The guard's report deletes the central gateway row, but pods whose persistent disk still holds the legacy `credentials.json` entry re-adopt it into the gateway on their next not-connected 404 (`RemoteCredentialStore.adoptFallback`). In a multi-agent workspace, each waking pod re-seeded the row its siblings had just deleted — workspace-wide whack-a-mole with a fresh Sentry error per round. This is where the wrong credential type kept re-entering the pipeline; no current client save path can create these rows (pasted keys are live-verified since HOU-1107).
2. **Per-sync refire on a failed report.** A failed dead-row report deleted the guard's dedupe entry, so the next serve sync (every turn and hydrating route) re-emitted the `console.error` — one Sentry event per sync while the control plane refused the delete. The host route also let a gateway failure escape as an unhandled 500.

## Fix

- **`@houston/protocol/google-key`** — shared `deadGoogleApiKey` predicate (google + `api_key` + not `AIza…`), used by both ends of the pipeline. Light subpath export so serve's import chain stays free of the zod-heavy barrel (the barrel import pushed `pi-providers.test.ts`'s import hook past its 10s budget).
- **Remote store (host)** — never adopt or `PUT` a dead google key centrally. A legacy fallback row is dropped from the pod's disk (with a visible warn) instead of resurrected; `putRemote` refuses as the last exit toward the gateway.
- **Capture (host)** — a google `api_key` export that is OAuth material is rejected with a clean 400 before storing, so a connect can't seed the central store with a key every serve refuses.
- **Serve guard (runtime)** — the error event fires ONCE per (scope, provider, token) per pod lifetime; failed reports retry on later syncs quietly (warnings), and an in-flight report is never duplicated.
- **`/sandbox/credential/revoked` (host)** — a store/gateway failure answers a clean 502 for the runtime to retry instead of an unhandled-throw 500.

## Expected effect

Each affected pod fires at most one more event, its report deletes the central row, and its own disk copy can no longer resurrect it — the fleet converges to quiet as pods wake. New bad rows cannot enter from any path in this repo. If the gateway itself persistently refuses a digest-delete (not observable from this repo), the residual signal is one error per pod wake, no longer one per turn.

## Tests

- runtime: failed report logs one error, retries quietly, stops once delivered; existing refuse/report/dedupe tests unchanged and green
- host: dead-key adoption is refused and the fallback row dropped; `put` refuses a dead google key; a real `AIza…` key still adopts/stores/captures; capture rejects an OAuth-type export; revoked route answers 502 on store failure
- Full `@houston/protocol` / `@houston/domain` / `@houston/runtime` / `@houston/host` suites green (2,705 tests); `pnpm check` and `pnpm check:boundaries` clean; tsgo typecheck clean

### DIFF
--- a/packages/host/src/channel/capture-credential.test.ts
+++ b/packages/host/src/channel/capture-credential.test.ts
@@ -73,3 +73,64 @@ test("heal mode rejects an api-key export before storing", async () => {
   expect(result.ok).toBe(false);
   expect(puts).toBe(0);
 });
+
+test("a google api-key export that is an OAuth-type token is rejected before storing", async () => {
+  // The capture-side door of Sentry HOUSTON-APP-567: pushing a legacy runtime
+  // entry centrally would seed the store with a key every serve refuses.
+  let puts = 0;
+  vi.stubGlobal(
+    "fetch",
+    vi.fn(async () =>
+      Response.json({
+        provider: "google",
+        kind: "api_key",
+        key: "ya29.a0LegacyOAuthToken",
+      }),
+    ),
+  );
+  const result = await captureRuntimeCredential({
+    endpoint: { baseUrl: "http://runtime", token: "runtime-token" },
+    credentials: {
+      get: async () => null,
+      put: async () => {
+        puts++;
+      },
+      remove: async () => {},
+      removeIfAccess: async () => false,
+    },
+    workspaceId: "workspace",
+    provider: "google",
+  });
+  expect(result).toMatchObject({ ok: false, status: 400 });
+  expect(result.ok === false && result.error).toMatch(/OAuth-type token/);
+  expect(puts).toBe(0);
+});
+
+test("a real AIza google api-key export still captures", async () => {
+  const stored: WorkspaceCredential[] = [];
+  vi.stubGlobal(
+    "fetch",
+    vi.fn(async () =>
+      Response.json({
+        provider: "google",
+        kind: "api_key",
+        key: "AIzaSyRealKey",
+      }),
+    ),
+  );
+  const result = await captureRuntimeCredential({
+    endpoint: { baseUrl: "http://runtime", token: "runtime-token" },
+    credentials: {
+      get: async () => null,
+      put: async (credential) => {
+        stored.push(credential);
+      },
+      remove: async () => {},
+      removeIfAccess: async () => false,
+    },
+    workspaceId: "workspace",
+    provider: "google",
+  });
+  expect(result).toEqual({ ok: true, provider: "google" });
+  expect(stored[0]?.accessToken).toBe("AIzaSyRealKey");
+});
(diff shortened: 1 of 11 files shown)


cravenceiling:
Conflicts resolved against main (now carrying #1293 + #1297) in 759352bb, and the combined behavior verified.

**The conflict** was test-file-only: #1297 added its tombstone tests at the same spot this branch added the store-failure 502 test. Both sets kept; they share the merged `post()` helper (which threads an injectable `RevocationTombstones`).

**Semantic review of the auto-merged production files** (the part git can't check):
- `credential-revoked.ts`: this PR's store-failure catch sits UPSTREAM of #1297's tombstone mark — a gateway hiccup answers 502 and neither marks a tombstone nor fires the "refilled and revoked AGAIN" escalation. Only a CONFIRMED delete marks. Correct composition.
- `remote-store.ts adoptFallback`: #1297's tombstone check runs first (team-scope key, matching the legacy fallback's scope), then this PR's dead-google-key drop. One benign quirk: while a tombstone is live the dead file entry isn't dropped that round — it's dropped on the first adoption attempt after the TTL. No loop either way, since `putRemote` refuses dead keys as the last exit.
- `capture-credential.ts` × #1297's clear-on-capture: a dead google export returns `ok:false`, and the `revocations.clear` in `ProxyChannel.captureCredential` is gated on `result.ok` — so a dead credential can never lift a tombstone. A pasted REAL `AIza…` key (live-verified) still clears it, so recovery after a real reconnect is immediate.
- `served-key-guard.ts` (runtime): the announced/pending/delivered split composes with #1297's `report-revoked.ts` dedupe — different report paths, same one-error-per-token-per-pod outcome; failed reports retry quietly on both.

**One cross-PR note:** the serve guard's dead-key reports flow through the same `/sandbox/credential/revoked` route that #1297 tombstones. A confirmed google dead-row delete therefore tombstones (workspace, team, google) for 15 min — which is exactly right: it blocks sibling pods' adoption during the whack-a-mole window this PR describes, and the two fixes reinforce each other.

Gates: protocol 29 / domain 193 / runtime 1073 / host 1434 tests green; Biome, typecheck, boundaries clean. MERGEABLE.

PR #1307, opened 2026-08-12
Serve/connect hygiene: api-key rollback, unconditional ghost clear, probe retry (PRODUCT-1321)

cravenceiling:
Fixes PRODUCT-1321, PRODUCT-1323, PRODUCT-1324. Should end Sentry **HOUSTON-APP-4YV** (~1,000 chronic serve-probe errors).

All three live in the serve/connect path (`packages/runtime/src/auth/serve*`, `packages/host/src/channel/proxy.ts`).

## L6 / PRODUCT-1321 (high): failed API-key connect leaves a working-but-doomed key

**Root cause.** `saveApiKeyCredential` (host `channel/proxy.ts`) POSTs the pasted key to the runtime, which **verifies AND persists** it into `auth.json` (`transport/provider-routes.ts` `handleApiKey`), and only then PUTs it centrally. If the central PUT fails, the route reports a failed connect — but the runtime keeps a usable key that the central store never learned about. That key is absent from `served-providers.json`, so the serve sync's provenance gate means an authoritative central 404 can never remove it. It works until the pod recycles, then vanishes with the emptyDir: the "spontaneous disconnect".

**Fix.** On a central PUT failure the host now rolls the runtime entry back via a new narrow runtime route, `DELETE /auth/:provider/api-key` (`transport/api-key-rollback.ts`). The rollback is **key-matched**: the runtime removes the entry only if it still holds that exact key, so a concurrent connect's newer key survives. A rollback transport failure is logged loudly but never masks the central failure — the user keeps seeing the real connect error either way, and a failed connect now leaves no local residue.

**Alternative considered and rejected:** reordering to central-first ("store centrally, hydrate via serve"). The runtime's live verification (`verifyApiKey`) needs the key runtime-side *before* anything is stored ("connected" must mean the key works), so central-first requires a new verify-without-persist runtime route plus reshaping the connect flow's immediate-connected signal (today the runtime write is what flips status instantly). That's the cleaner long-term shape, but it's a larger contract change for the same end state; the compensating rollback is minimal-risk and converges to central truth in every failure interleaving.

## L8 / PRODUCT-1323 (medium): ghost `.credentials.json` survives outside the manifest gate

**Root cause.** PR #1303's `clearGhostClaudeCredential()` ran only inside `if (manifest.has(probe.id))`. A ghost materialized SDK credential planted while anthropic was **never** in the served manifest (a setup pod's non-attributed connect whose row was later removed; a self-host row verify-rejected before any successful serve) was never cleared — and the provenance gate then blocks the revocation reporter, so every turn burns on the dead family until the file's token expires. PRODUCT-1307 through a different door.

**Fix.** The clear now runs on **every** authoritative anthropic `not-connected` probe, outside the manifest gate. The function was already serve-mode-reached-only, personal-scope-guarded (HOU-976: a member's verdict can't delete the team's file), and a no-op without the file. The manifest-gated `auth.json` removal is unchanged.

## L9 / PRODUCT-1324 (medium): ~40 unretried concurrent probes, ERROR per blip, provider silently un-applied

**Root cause.** `runServedSync` fired ~40 concurrent probes with no retry on every turn start and several hydrating routes; one transient socket failure produced a Sentry ERROR per provider (HOUSTON-APP-4YV) **and** silently un-applied that provider for the sync — a freshly recycled pod's first turn could read a connected provider as not-connected. The log detail kept only `err.message` ("fetch failed"), losing the undici `cause` code.

**Fix** (probe logic extracted to `auth/serve-probe.ts`, which also brings `serve.ts` back down in size):
- **One retry after 250ms**: the first failure is a WARN breadcrumb; the error becomes final (and Sentry-visible via `serve-log.ts`'s existing transition dedup) only when the retry fails too. A one-off blip never alerts; a persistent gateway outage still logs exactly one ERROR per provider per detail-transition. A *timed-out* first attempt is final immediately — it already stalled 10s, a 250ms-later second 10s stall wouldn't heal it, and the next sync re-probes anyway.
- **Concurrency capped at 8** (tiny inline worker pool, no dependency) — the 40-socket burst from a just-woken pod was itself a plausible source of the transients.
- **Cause codes in the detail**: `describeProbeError` walks `cause` chains (incl. AggregateError / errno) so the log and Sentry event say `fetch failed (cause: ECONNRESET)` instead of `fetch failed`.
- **Failure semantics unchanged**: a probe that fails both attempts removes nothing (credential + manifest entry survive) and is re-attempted on the next sync.

## Tests

- **L6**: runtime route test (exact-key rollback, wrong-key untouched, keyless 400, idempotent no-op); host contract tests (central PUT failure → rollback DELETE with the exact key + central error rethrown; failing rollback logged, never masks; successful connect fires no rollback).
- **L8**: never-manifested ghost cleared on authoritative not-connected; personal scope still never touches the pod-shared file (manifest or not).
- **L9**: unit tests for retry/timeout-skip/pool-order/concurrency-cap/cause-code extraction; full-sync tests — one transient failure retries, provider applies, zero ERRORs; both-attempts-fail logs ONE error carrying ECONNRESET, removes nothing, re-attempts next sync with the repeat deduped to WARN.

`pnpm --filter @houston/runtime test` (118 files, 1091 tests) and `pnpm --filter @houston/host test` (155 files, 1437 tests) pass; `pnpm check` (Biome) clean; `tsc --noEmit` clean in both packages.

### DIFF
--- a/packages/host/src/channel/contract.test.ts
+++ b/packages/host/src/channel/contract.test.ts
@@ -1,6 +1,6 @@
 import { createServer, type Server } from "node:http";
 import type { AddressInfo } from "node:net";
-import { afterAll, beforeAll, describe, expect, test } from "vitest";
+import { afterAll, beforeAll, describe, expect, test, vi } from "vitest";
 import { MemoryCredentialStore } from "../credentials/store";
 import type { Agent, Workspace } from "../domain/types";
 import { FakeLauncher } from "../launcher/fake";
@@ -194,6 +194,10 @@ let proxyCustomEndpointBody: unknown = null;
 let proxyClaudeOAuthBody: unknown = null;
 /** When set, the fake runtime rejects /auth/:provider/api-key with this reply. */
 let proxyApiKeyRejection: { status: number; body: unknown } | null = null;
+/** DELETE /auth/:provider/api-key rollbacks the fake runtime received. */
+let proxyApiKeyRollbacks: { provider: string; body: unknown }[] = [];
+/** When true, the fake runtime fails the rollback DELETE with a 500. */
+let proxyApiKeyRollbackFails = false;
 let proxyScrubFailuresRemaining = 0;
 let proxyScrubCalls = 0;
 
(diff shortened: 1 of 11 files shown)


cravenceiling:
Review fix pushed (5fa112da): the unconditional ghost clear would have deleted the DESKTOP/self-host browser login's legitimate `.credentials.json` — those hosts answer every anthropic serve with the marked 404 ("not served on this deployment"), and the manifest gate this PR removed was what kept that answer from deleting local files. The refusal now carries a second marker (`x-houston-not-served-here`) and the ghost clear only runs on a real workspace-disconnect 404. Regression tests added on both sides (runtime + host). Full suites green (1092 runtime / 1441 host).

PR #1308, opened 2026-08-12
Never refresh with an empty token: serve re-sync guard + serve-skew margin (PRODUCT-1317)

cravenceiling:
## Root cause

Gate #2 stores a served credential as `{type:"oauth", access, refresh:"", expires}` (`packages/runtime/src/auth/auth-file.ts`). pi-ai 0.84.1 does not know that shape: `resolveStoredOAuth` (`dist/auth/resolve.js`) routes ANY stored OAuth entry within 5 minutes of expiry (`DEFAULT_OAUTH_MINIMUM_VALIDITY_MS`) through `oauth.refresh(current)`, and every provider refresher POSTs `credential.refresh` verbatim to its token endpoint. The gateway serves with exactly 5 minutes of margin (zero slack) and the desktop host's serve path only refreshed inside 2 minutes (`isExpiring` default skew), so pi could receive a token already inside its refresh window — and fired `POST refresh_token=""` at the provider (openai-codex), or minted with `Bearer ""` and 401'd into a spurious reconnect card (github-copilot). Seen live in production as **PRODUCT-1293**, which this root-causes.

## Fix

**Engine guard (Houston-owned seam, no pi patch)** — `packages/runtime/src/auth/empty-refresh-guard.ts`, wired into `HoustonAuthStore` (pi's `CredentialStore`):
- `read()`: an access-only entry already inside pi's 5-minute window triggers the single-flighted `syncServedCredential()` (bound from `serve.ts`; a direct import would cycle serve → storage → credential-store → serve), so a fresh central token lands **before** pi's expiry check runs. The gateway refreshes centrally when serving a sub-5-minute token, so the re-serve genuinely closes the race.
- `modify()`: pi's refresh closures dereference `current` and hand it to the provider refresher — an access-only (`refresh:""`) entry is masked to `undefined`, which takes pi's own "logged out meanwhile" branch: the entry is left unchanged and pi serves the stored access token as-is via `toAuth` (side-effect-free per its `OAuthAuth` contract). The token's remaining validity still gets used; once truly expired the request 401s into the typed `unauthenticated`/`token_expired` card. **No refresh POST ever carries the empty string.**

**Desktop/self-host serve margin** — `packages/host/src/routes/credential.ts`: the serve-time refresh now fires below 6 minutes of validity (pi's 5-minute floor plus a minute of slack), passed as an explicit `skewMs` at the serve sites — `isExpiring`'s 2-minute default is untouched for its other callers. The anthropic staleness gate uses the same margin (a served anthropic entry also lands access-only in auth.json where pi's usage probes resolve it). The refresh coalescer threads the caller's margin through its cached-result and in-flight re-checks so it can't hand back a token the route already judged too short-lived.

## Tests

- `packages/runtime/src/auth/empty-refresh-guard.test.ts` — pins the store-level contract with pi's exact refresh-closure shape: an expiring `refresh:""` entry re-syncs at read and the refresher is **never** invoked (sync healthy, sync down, and sync unbound); a refresh-bearing credential still refreshes through `modify`; long-lived served tokens and pasted no-expiry tokens never trigger the sync; window-edge cases for `needsServeSync`.
- `packages/host/src/routes/credential.test.ts` — a token with 4 minutes of validity (fine for the old skew, inside pi's window) is refreshed at serve time; a healthy long-lived token serves with zero token-endpoint calls; an anthropic token below the floor answers the marked 404.
- Full suites green: `@houston/runtime` 118 files / 1086 tests, `@houston/host` 155 files / 1437 tests; `tsc --noEmit` clean for both; `pnpm check` (Biome) clean.

Fixes PRODUCT-1317. Root-causes PRODUCT-1293.

### DIFF
--- a/packages/host/src/credentials/refresh-coalescer.ts
+++ b/packages/host/src/credentials/refresh-coalescer.ts
@@ -40,6 +40,13 @@ export type CredentialRefreshRun = {
    * keeping a private one.
    */
   doRefresh?: CredentialRefresher;
+  /**
+   * Expiry margin THIS caller needs (`isExpiring`'s `skewMs`). The serve route
+   * passes its pi-validity-floor margin (routes/credential.ts, PRODUCT-1317) so
+   * the in-flight re-check can't hand back a token the route already judged too
+   * short-lived. Absent = `isExpiring`'s default.
+   */
+  skewMs?: number;
 };
 
 /**
@@ -96,7 +103,7 @@ export class CredentialRefreshCoalescer {
     if (cached) {
       if (
         this.now() - cached.at < REFRESH_RESULT_TTL_MS &&
-        !isExpiring(cached.cred)
+        !isExpiring(cached.cred, args.skewMs)
       )
         return Promise.resolve(cached.cred);
       this.results.delete(key);
@@ -153,7 +160,7 @@ export class CredentialRefreshCoalescer {
     // it and persisting the result — would recreate the row they deleted and
     // hand their agents a live token for a revoked connection.
     if (loaded === null) throw new CredentialGoneError(args.provider);
-    if (!isExpiring(loaded)) {
+    if (!isExpiring(loaded, args.skewMs)) {
       this.remember(key, loaded);
       return loaded;
     }
--- a/packages/host/src/routes/credential.test.ts
+++ b/packages/host/src/routes/credential.test.ts
@@ -622,6 +622,30 @@ test("an expiring anthropic credential whose refresh fails degrades to marked 40
   expect(r.out.headers?.["x-houston-not-connected"]).toBe("1");
 });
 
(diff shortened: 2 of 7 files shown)


cravenceiling:
Review fix pushed (319398fb): the anthropic STALE refusal must keep `isExpiring`'s default margin. Judged at 6 minutes it would refuse healthy 5-6-minute tokens until every gateway rolls the 10-minute ServeSkew — and the marked 404 is an authoritative disconnect (the runtime deletes its served entry and, since PRODUCT-1323, the ghost file): an org-wide anthropic flap once per token lifetime, coupled to deploy order. The 6-minute margin still drives the serve-time refresh; short-but-live tokens are served and the runtime's empty-refresh guard owns pi's window. Tests adjusted + deploy-order regression test added. Full suites green (1086 runtime / 1441 host).

Issue #238, opened 2026-05-22
always-on Dockerfile fails to build + crashes on first boot (3 bugs)


cravenceiling:
The `always-on/Dockerfile` (documented as shipped in `knowledge-base/architecture.md:113`) fails three independent ways when followed end-to-end.

## 1. Workspace fails to load during `cargo build`

```
error: failed to load manifest for workspace member `/src/app/houston-tauri`
referenced by workspace at `/src/Cargo.toml`
Caused by: No such file or directory (os error 2)
```

`Cargo.toml` lists `app/houston-tauri` and `app/src-tauri` as workspace members (lines 6 + 22). Cargo refuses to load the workspace if any member's manifest is missing — even when `-p houston-engine-server` excludes them from the build. The Dockerfile only copies `engine/`.

## 2. `houston-agent-files` fails to compile

```
error: couldn't read `engine/houston-agent-files/src/../../../ui/agent-schemas/src/activity.schema.json`: No such file or directory
```

`engine/houston-agent-files/src/schemas.rs:3-7` embeds 5 JSON schemas from `ui/agent-schemas/src/*.schema.json` at compile time via `include_str!`. The Dockerfile doesn't copy that tree.

## 3. Container crash-loops on first boot

After the build succeeds, the runtime container exits with code 101 in a restart loop:

```
thread 'main' (1) panicked at engine/houston-engine-server/src/main.rs:131:10:
engine state init failed: Failed to open engine DB

Caused by:
    0: Failed to create data directory
    1: Permission denied (os error 13)
```

The Dockerfile creates user `houston` with `--create-home --home-dir /data`, but neither `/data/.houston` nor `/data/Houston` exists in the image. When `docker-compose` mounts a named volume at `/data/.houston`, Docker creates the path as root before the container starts, so the unprivileged `houston` user can't write the libSQL DB.

## Repro

```bash
docker build -t houston/engine:dev -f always-on/Dockerfile .   # fails at #1
# fix #1, retry → fails at #2
# fix #2, retry → build succeeds, container crash-loops with #3
```

## Fix

PR copies the two missing workspace-member dirs (`app/houston-tauri`, `app/src-tauri`) and the `ui/agent-schemas/` tree into the build stage, and pre-creates + chowns the two data dirs (`/data/.houston`, `/data/Houston`) before `USER houston` so volume mounts inherit correct ownership on first attach.

## Environment

- Host: WSL2 (Ubuntu) + Docker Desktop
- Reproduced from `main` at 3cac688

PR #1312, opened 2026-08-12
feat: fenced writes — boot lease claim + generation-aware sync (PRODUCT-1303)

cravenceiling:
Engine half of Phase 2 of the stateless architecture (PRODUCT-1303) — companion to gethouston/cloud#238 (the gateway/pod-store half). **Deploy-skew safe in both directions**: against an old gateway (no fencing token, no generations) this engine sends no fencing headers and no preconditions — byte-identical wire behavior to today. Merge order does not matter.

## What's here
- **Boot-time lease claim** (managed-store-config, before hydrate): the engine claims the agent's write lease for THIS boot and seeds the fence holder. Covers container restarts and node reschedules (no control-plane wake involved); 404 = old/unfenced gateway → proceed exactly as today.
- **http-store**: captures `X-Houston-Fencing-Token` from store responses and echoes it (+ `X-Houston-Boot-Id`) on writes once seen; parses per-object generations; guarded writes never retry on 502/503/504 (a lost success would 412 its own retry); typed `StoreFencedError` (409) / `StoreConflictError` (412) beside the existing 413.
- **hydrate/syncBack**: manifest entries are `{hash, generation?}`; per-object CAS — known objects write with their hydrated generation, new files create-only; a 412 refreshes the remote manifest once (memoized per pass) and retries that object once, then records it in `SyncResult.conflicts`; deletes are conditional and never remove an object someone else rewrote; a 409 aborts the pass and propagates.
- **daemon**: a fence rejection latches the daemon — one err-level log, watcher/timers stopped, and `stop()` skips the shutdown drain (a fenced pod dies quietly instead of fighting the fence).
- **shared-mirror**: the org-shared prefix (multi-writer by design) gets generation preconditions — its advisory conflict detection becomes enforced through the existing `onConflict` callback — but never fencing.
- Generation `0` normalizes to 'no generation' (the dir backend has none; GCS never mints 0).

## Validation
162 runtime-client tests + host store-sync + a real-HTTP fenced-takeover e2e; plus the cross-repo ladder (real pod-store + Postgres + SIGSTOP zombie, CAS against real GCS generations, human desktop test) — detail in PRODUCT-1303.

Port note: the `ObjectStore` widening is optional/additive — every existing adapter and inline test fake compiles unchanged.


cravenceiling:
Consolidated into the phases 2+3 PR per review preference (one PR per repo).

Issue #241, opened 2026-05-22
[BUG]: Can't complete tutorial after creating skill


cravenceiling:
After telling the agent to create a skill based on the previous it's not possible to continue sending messages. Neither the user's nor the agent's messages will appear in the chat from here

PR #1315, opened 2026-08-12
feat: phase 3C — transcript/doc shadow-write + turnlog capture (PRODUCT-1303)

cravenceiling:
**Engine slice (houston)** — stacked on #1312 (fenced writes); retargets `main` when it merges. Companion to the cloud stack gethouston/cloud#239–#241. The conversation FILE stays authoritative; every new path is best-effort shadow that can never block, delay, or fail a turn. Behind HOUSTON_TRANSCRIPT_DUAL_WRITE / HOUSTON_TURN_LOG; unset ⇒ nothing constructed, byte-identical.

- Transcript shadow: atomic file write+rename FIRST (ordering unchanged), then a detached per-conversation queue POSTs a DEEP-CLONED post-write snapshot (closes the parse-cache aliasing hazard) through a host /sandbox facade → pod-store, with the phase-2 fencing headers and the expectedCount divergence etag. Failures mark the conversation dirty; the next op repairs via the idempotent repair route.
- Doc shadow: the five .houston families keep their file RMW (files remain the agent-readable surface); each write is followed by a best-effort If-Match PUT. The FS watcher also shadows the agent's prompt-sanctioned direct learnings.json write.
- Turnlog: FrameForwarder batches frames (250ms/32, terminal flushed immediately) carrying the producer's authoritative seq; StandingFrameCapture pumps the standing runtime's SSE and is triggered at turn dispatch AND routine fire — turns with NO client are captured (the case a tee design misses), with an explicit test.
- Reconcile reads via the gateway reply-after route under dual-write, file fallback on any error. File reads and the duplicate conversationKey builders are retired in a later PR, not here.

Boundary check green (no pg/redis in packages/**); 170 host + 38 runtime-store tests.

### DIFF
--- a/knowledge-base/files-first.md
+++ b/knowledge-base/files-first.md
@@ -23,6 +23,23 @@ Houston uses files, not a DB, for everything. The only SQLite left is a read-onl
 > as `ConversationsChanged`, which the gateway fans to every client so an open
 > chat on another device reloads the same transcript.
 
+### Managed durable-turn shadows
+
+Two managed-cloud rollout caps mirror this file model without changing its
+authority. `HOUSTON_TRANSCRIPT_DUAL_WRITE=1` queues transcript mutations only
+after the atomic file rename and projects watcher-observed `activity`,
+`routines`, `routine_runs`, `config`, and `learnings` files to the gateway.
+Failures, revision conflicts, and deploy skew are contained in the shadow path;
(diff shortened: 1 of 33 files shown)


cravenceiling:
Consolidated into the phases 2+3 PR per review preference (one PR per repo).

PR #1338, opened 2026-08-18
Latch the runtime launcher closed on host shutdown (PRODUCT-1399)

cravenceiling:
## Summary

Sentry HOUSTON-APP-4YV (`[serve] credential <id>: fetch failed (cause: ECONNREFUSED)`) — PRODUCT-1399.

The original root cause of this Sentry group (probe fan-out with no retry, PRODUCT-1324) is already fixed and deployed (release `b9c4c6da`, ~Aug 14). What is left is a small, exact residual: **41 events per occurrence, one pod per day**, all from the same mechanism, which pod logs pin down:

1. Pod termination (k8s `ScaleDown` / `Killing`) → `host.stop()` → `launcher.shutdownAllAndWait()` SIGTERMs the runtime and **clears the live-set before the child has exited**.
2. A dispatch lands in that 3–5 s drain window (the client's `/providers` + `/auth/status` poll every 15–60 s, an SSE resume, a routine): `ensureAwake` finds no running entry and **respawns a fresh runtime**.
3. The host closes its listener; the new runtime's first serve sync hits `ECONNREFUSED` for all 41 known providers → 41 error events, and the orphaned child keeps the dying pod's in-flight request alive.

On the desktop the same race spawns a runtime the exiting host never kills (an orphan holding the agent dir + a port).

## Fix

- `ProcessLauncher` latches `closed` on `shutdownAll` / `shutdownAllAndWait`; `ensureAwake` refuses with a typed `LauncherClosedError` (re-checked after the port-allocation await, the one gap where a boot entered before shutdown could still spawn past the sweep).
- `wakeForDispatch` maps `LauncherClosedError` → **503 + Retry-After** so the client retries against the replacement pod / next app start instead of a generic 500.

Host-only, ~30 lines, no wire/protocol change. Rides the next regular engine roll — no dedicated roll needed.

## Tests

- drain-window `ensureAwake` is refused, never respawned; stays refused after the drain
- sync `shutdownAll` latches the same way
- a boot parked on `allocatePort` when shutdown starts is refused (no child past the sweep)
- probe route and non-probe route answer 503 for a closed launcher

`pnpm --filter @houston/host test` (1513 pass), typecheck, `pnpm check`, `pnpm check:boundaries` all green.

## Before / after

**Before:** in Cloud Logging for any pod evicted while a client polls (`resource.labels.pod_name=<pod>`), the sequence `runtime shutdown requested signal=SIGTERM` → ~3–5 s later a NEW `runtime listening` → `[serve] credential <id>: fetch failed (cause: ECONNREFUSED)` × 41 → `[serve] applied central credentials: (none)`. Sentry 4YV gains 41 events per such pod.

**After:** the same eviction shows SIGTERM → no second `runtime listening`; the in-flight poll gets a `503 {"error":"the host is shutting down; retry shortly"}` and the pod exits promptly. Sentry 4YV's ECONNREFUSED family stops growing on the rolled release (watch `release:engine-pod@<sha>` after the roll).

### DIFF
--- a/packages/host/src/channel/probe-wake.ts
+++ b/packages/host/src/channel/probe-wake.ts
@@ -1,5 +1,10 @@
 import type { ServerResponse } from "node:http";
-import type { ChannelCtx, RuntimeEndpoint, RuntimeLauncher } from "../ports";
+import {
+  type ChannelCtx,
+  LauncherClosedError,
+  type RuntimeEndpoint,
+  type RuntimeLauncher,
+} from "../ports";
 import { json } from "../routes/http";
 
 /**
@@ -53,7 +58,8 @@ export async function wakeForDispatch(
   res: ServerResponse,
 ): Promise<RuntimeEndpoint | null> {
   const wake = launcher.ensureAwake(ctx.agent);
-  if (method !== "GET" || !PROBE_ROUTES.has(rest)) return wake;
+  if (method !== "GET" || !PROBE_ROUTES.has(rest))
+    return wake.catch((err) => rejectIfClosed(err, res));
 
   // A boot that fails after this request has been answered has no one left to
   // tell; the next probe re-triggers the spawn and surfaces the failure then.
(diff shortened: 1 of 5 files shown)


cravenceiling:
Complementary runtime-side half in #1339: a sweep in which every probe fails identically is logged once (`[serve] control plane unreachable for all N providers: …`) instead of 41 times. Disjoint files; either order merges.

PR #1353, opened 2026-08-19
fix(shell): normalize tracing-log targets so noisy-dep Sentry demotes actually work (PRODUCT-1437)

cravenceiling:
## Root cause

Sentry HOUSTON-APP-S (`failed to check for updates: …/latest.json`, 3.8k events since May, still emitting from 0.5.47 and 0.6.x) survived the HOU-1104 demote because of a `tracing-log` subtlety: `log`-crate records reach tracing through `LogTracer`'s static callsites, whose `Metadata::target()` is literally `"log"` — the real target only exists on the *normalized* metadata. The `sentry_tracing::layer().event_filter(...)` only receives the un-normalized callsite metadata, so the target-based demote list (`tauri_plugin_updater`, and `rustls_platform_verifier` from PRODUCT-1430) never matched any log-crate source. Every failed 5-minute update poll still became a full Sentry event. The existing unit tests passed because they called the demote predicate directly with the real target string, bypassing the bridge.

## Fix

- Replace the `event_filter` with an `event_mapper` that normalizes log-bridged metadata (`tracing_log::NormalizeEvent::normalized_metadata()`) before applying the demote predicate, mirroring the default breadcrumb/event mapping otherwise.
- Move the decision into a dedicated `app/src-tauri/src/sentry_events.rs` module (keeps `logging.rs` under the line cap).
- Pipeline-level tests: emit real `log::error!` records through `LogTracer` into the layer and assert what reaches Sentry via the in-memory test transport (`sentry` `test` feature), instead of unit-testing the predicate. Demoted targets produce no event; other `log`-crate errors and `tracing::error!` from app code still do.

This also makes the PRODUCT-1430 `rustls_platform_verifier` demote effective for the first time — that crate logs via the `log` crate too.

## Before the fix (reproduction)

1. On this branch, revert `sentry_events::layer()` to the old `event_filter` closure.
2. `cd app/src-tauri && cargo test --lib sentry_events` → `log_bridged_updater_error_stays_a_breadcrumb` **fails**: the log-bridged updater error is captured as a full Sentry event (exactly what HOUSTON-APP-S shows in production).
3. Equivalent manual repro: run a packaged build with no network route to github.com; each 5-minute updater poll produces one HOUSTON-APP-S event.

## After the fix (verification)

1. `cd app/src-tauri && cargo test --lib` → 224 passed, including the four pipeline tests in `sentry_events`.
2. `cargo check` clean.
3. Post-release: HOUSTON-APP-S (and HOUSTON-APP-PE) event counts for the new release version should drop to zero in Sentry; the demoted lines still appear as breadcrumbs on real crash events and in `backend.log`.

## Notes

- No server-side alternative exists on the current Sentry plan (message-based inbound filters and delete-and-discard are Business-plan features), so old clients keep emitting until they update; a stuck cohort on 0.4.26 that cannot reach github.com will keep emitting regardless.

PRODUCT-1437

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2084,6 +2084,7 @@ dependencies = [
  "hyper",
  "hyper-util",
  "keyring",
+ "log",
  "notify-rust",
  "objc2",
  "objc2-app-kit",
@@ -2107,6 +2108,7 @@ dependencies = [
  "tokio",
  "tracing",
  "tracing-appender",
+ "tracing-log",
  "tracing-subscriber",
  "windows-sys 0.59.0",
 ]
--- a/app/src-tauri/Cargo.toml
+++ b/app/src-tauri/Cargo.toml
@@ -39,6 +39,10 @@ sentry = "0.42"
 # Layer is a no-op until sentry::init has run, so safe to register
 # unconditionally in logging.rs even when SENTRY_DSN is empty.
 sentry-tracing = "0.42"
+# `NormalizeEvent` in sentry_events.rs: log-crate records reach tracing with
+# callsite target "log"; the real target (needed to demote known noisy deps)
+# only exists on the normalized metadata.
+tracing-log = "0.2"
 # keyring v3 backends are OPT-IN per platform. Without a Linux feature the
 # crate silently compiles its in-memory mock on Linux — auth_set_item
 # "succeeds" into a throwaway object and every session evaporates on quit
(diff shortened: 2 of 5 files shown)


cravenceiling:
## Status: parked, desktop is frozen

The local desktop app no longer ships updates (only cloud does), so this change reaches no existing user. It stays open as insurance for `main` only: the HOU-1104 / PRODUCT-1430 demotes on `main` silently do nothing for log-crate sources, and the pipeline tests here encode that gotcha for any future desktop build.

The operational fix for HOUSTON-APP-S is Sentry key rotation, no release needed. The org has one project and one client key ("Default") shared by every surface; desktop binaries bake the DSN at build time, while cloud pods read `SENTRY_DSN` at runtime and cloud web re-bakes it on each deploy. Plan: create a second key for cloud, repoint pods (env) and web (CI secret), then disable or rate-limit the old key so frozen desktop binaries are rejected at ingest (dropped events do not count toward quota). Tradeoff: that silences all desktop telemetry, not just the updater noise.

Tracked in PRODUCT-1437 (Todo).

PR #1376, opened 2026-08-21
docs(kb): audit the agent workspace for stateless cloud (PRODUCT-1473)

cravenceiling:
## Summary

Adds `knowledge-base/houston-folder/`: one row per path under an agent's workspace (`.houston/**`, `.agents/skills`, context files, uploads, user files, `.houston/runtime/**`, root-level objects) with:

`source of truth · read by pool worker from · written by pool worker to · projected doc family · served to the app from (asleep) · claim scope`

plus `divergences.md` (28 places two copies can disagree and what reconciles each, or nothing) and `wake-writers.md` (what a standing pod rewrites on boot or wake, in order). Every claim cites the module that makes it true.

`CLAUDE.md` previously said there is no separate knowledge base; it now points at `knowledge-base/` as code-cited audits where the code wins on disagreement.

Linear: PRODUCT-1473

## Highest-value findings for the follow-on work

- A claimed pool turn hydrates everything but may land only its conversation, its sessions dir, `activity.json` and `routine_runs.json`; every other file the model writes is dropped at sync-back with one `console.info` (D1).
- A settings op emits no event, so the `providers` view keeps the captured `activeModel` until the next wake (D2). `settings.json`, `custom-endpoint.json`, `qwen-region.json`, `auth.json`, `store-publication` and schema re-seed emit nothing anywhere, so SSE fan-out alone will not surface them.
- The worker already returns op event types; the gateway reply struct discards them.
- `runtime.log` lives inside the data dir, is excluded by nothing, and is hydrated onto every pool turn (D17).
- The 412 retry re-issues at the refreshed generation: generation preconditions detect concurrent writes, they do not protect (README, D15, D23).

## Verification

Docs only; no code paths change. Biome has no inputs here.

Before: `ls knowledge-base` fails; the per-path truth is spread across `object-sync/hydrate.ts`, `turn-filesystem.ts`, `op-apply.ts`, `docs/projector.ts`, the gateway `classifyPoolOp` and `dispatch()` chain.

After: `knowledge-base/houston-folder/paths.md` answers each column for each path; spot-check any cell against the cited module.

## Review

Two independent adversarial passes (Claude, Codex) against both repos; all confirmed findings folded in (commit 2). Dismissed: the suggestion to strip Postgres table and gateway function names. They are data-model vocabulary the audit needs; no hostnames, namespaces, secrets, ticket ids or private-repo links appear.


cravenceiling:
Withdrawn: the deliverable moves to a Linear document on the issue instead of a knowledge-base/ directory in the repo (owner decision).

PR #1397, opened 2026-08-25
fix(runtime): pool turns honor their pinned provider; post-failure serves bypass the negative cache (PRODUCT-1515)

cravenceiling:
Fixes PRODUCT-1515 — a chat asked to reconnect Anthropic while the conversation's selected provider was OpenRouter, and reconnecting "succeeded" only for the next turn to ask again.

## Root causes

**1. Wrong-provider resolution on the stateless (pool) turn path.** A pool turn ran on whatever credential the dispatcher attached: `TurnRequest` had no `provider` field, and `executeTurn` dropped the routine's own pinned provider (`RoutinePhase.provider` was computed by `prepareRoutineTurn` and never consumed). A dispatch that attaches a different provider's credential silently moved the turn — and its auth-error card — onto a provider the conversation never picked. The card then truthfully named that wrong provider ("Sign in to Anthropic again") while the picker showed OpenRouter.

**2. The reconnect loop.** The pod host's `RemoteCredentialStore` caches a "not connected" answer for 15s, and only its own `put()` invalidates it. A reconnect capture lands in the central store without passing through the pod's host process, so the auto-resume fired right after "Reconnected" re-read the cached null, got the marked 404 (which also drops the runtime's served entry), and re-rendered the reconnect card.

## Fix

- `TurnRequest.provider` (new, optional): the conversation's pinned provider, forwarded by the dispatcher. `piTurnRequest` runs the turn on the pin, falling back to the attached credential's provider only for legacy dispatches; `executeTurn` overlays a routine's pinned provider onto the turn. A mismatched credential now fails as the **pinned** provider's auth error instead of silently running (and billing) on the wrong one.
- The runtime's serve probe appends `fresh=1` while the provider's auth-failure mark is active; `handleSandboxCredential` then sheds the store's cached row (new optional `CredentialStore.invalidate`) before the read. The mark auto-heals on the credential change the fresh serve delivers, so the bypass lasts only while the failure does.

Dispatchers that today choose the served credential can start forwarding the conversation's pinned provider via the new field; absent it, behavior is unchanged.

## Repro (before)

1. Agent with a routine or task conversation pinned to OpenRouter (picker shows e.g. "OpenRouter: Fusion"); dispatcher attaches an Anthropic credential to the machine turn.
2. The turn runs on Anthropic; if that credential is dead, the chat shows "Sign in to Anthropic again" even though the conversation's provider is OpenRouter.
3. Reconnect Anthropic → "Reconnected" → auto-resume within 15s of the failing turn hits the host's cached "not connected" → the card returns.

## Verify (after)

- `pnpm --filter @houston/host --filter @houston/runtime --filter @houston/domain test` — green (new coverage: `pi-turn-request.test.ts`, pinned-provider e2e in `server.test.ts`, `types.test.ts` parse cases, `probePath` in `serve-probe.test.ts`, `invalidate` in `remote-store.test.ts`, `fresh=1` in `credential.test.ts`).
- A pool turn with `provider: "openrouter"` and an `anthropic` credential runs (and attributes any failure to) OpenRouter — see the `server.test.ts` case.
- A serve with `fresh=1` re-reads the central store inside the 15s cache window — see the `remote-store.test.ts` / `credential.test.ts` cases.

### DIFF
--- a/packages/host/src/credentials/remote-store.test.ts
+++ b/packages/host/src/credentials/remote-store.test.ts
@@ -107,6 +107,40 @@ test("404 means not connected and is cached as a negative result", async () => {
   expect(calls).toHaveLength(1);
 });
 
+test("invalidate sheds a cached negative so the next get sees a reconnect (PRODUCT-1515)", async () => {
+  // A reconnect capture lands centrally without passing through this process,
+  // so only an explicit invalidate can make the post-reconnect retry see it
+  // inside the 15s cache window.
+  const { calls, fetchImpl } = fakeFetch((_call, index) =>
+    index === 0
+      ? json({ error: "org not connected" }, 404)
+      : json(gatewayCredential()),
+  );
+  const s = store(fetchImpl);
+
+  expect(await s.get("ws_1", "openai-codex")).toBeNull();
+  s.invalidate("openai-codex");
+  const got = await s.get("ws_1", "openai-codex");
+  expect(got?.accessToken).toBe("AT-central");
+  expect(calls).toHaveLength(2);
+});
+
+test("invalidate is scoped to the acting identity", async () => {
+  const { calls, fetchImpl } = fakeFetch(() =>
+    json({ error: "org not connected" }, 404),
+  );
+  const s = store(fetchImpl);
+  const acting = { actingAs: actingAs() };
+
+  expect(await s.get("ws_1", "openai-codex")).toBeNull();
+  expect(await s.get("ws_1", "openai-codex", acting)).toBeNull();
+  // Invalidating the member's row leaves the team's cached answer in place.
+  s.invalidate("openai-codex", acting);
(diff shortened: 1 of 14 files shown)


cravenceiling:
Prod verification note (PRODUCT-1515): the reported user's session records show every failed turn during the incident window ran as `provider="openrouter", unauthenticated/no_credentials` — the runtime honored the conversation's pin throughout, and no turn ran on or was attributed to Anthropic server-side. The reported failures trace to a transient credential-serve outage plus the user's own exhausted provider credits; no pool-dispatched turn reached a real user org, so the pool-path provider bug fixed here is latent, not the live vector of that report.

Both fixes stand on their own: the pool path's provider selection defect is real in code (routine pin computed and dropped; `TurnRequest` had no provider field), and the 15s negative-cache race on the post-reconnect retry is real on the standing path. Details in the Linear thread.

PR #1399, opened 2026-08-25
feat(runtime): single-use pool workers unlock local code execution (PRODUCT-1454)

cravenceiling:
## Summary

Gives stateless pool turns a working shell with the least user-visible latency: the worker pod becomes **single-use** — it serves exactly one claimed turn and is replaced — which makes the pod single-tenant for its whole life and restores the standing pod's justification for in-container `bash`. Pods are replaced off the request path (warm spares), so the per-turn cost the user sees stays hydrate-only.

Pairs with a cloud-repo change that adds `runtimeClassName: kata-qemu` (VM isolation per pod, verified in staging), flips the pool to `HOUSTON_CODE_EXECUTION=local` + `HOUSTON_POOL_SINGLE_USE=1`, and recycles spent pods.

## Changes

- **`HOUSTON_POOL_SINGLE_USE=1`** (config): turn mode fails closed on `local` without single-use, and on single-use without `HOUSTON_TURN_CONCURRENCY=1`.
- **`turnCodeExecutionMode()`** (tool-selection): `local` is honored only on a single-use worker; a shared multi-turn worker degrades to `disabled` — one org's process tree, tmp residue, and env must never be readable by the next org's turn.
- **Spent latch** (`turn/single-use.ts`): marker on the pod volume written **before** the claimed turn runs, so a crash mid-turn fails closed. A restarted container in a spent pod idles unregistered and answers `503 worker_draining` until the orchestrator replaces the pod.
- **Server lifecycle**: a claimed `/turn` spends the worker (begin → turn → settled → graceful drain + exit 0). `/op` never spends it (host-side write, no model-directed code).
- **Sync-back scope widened**: `claimedTurnIncludes` now grants the agent's workspace **files**. Previously a spreadsheet the agent built or a doc it edited fell into `pool_writes_out_of_scope` and was silently dropped — this affected the clamped `write`/`edit` tools too, not just bash. `.houston/` internals beyond the explicit grants stay out of scope.
- **Prompt honesty**: the fallback system prompt no longer promises "run commands" when code execution is disabled.

## Testing

- `pnpm --filter @houston/runtime test` — 157 files / 1375 tests pass (new: single-use lifecycle, `turnCodeExecutionMode`, widened sync scope).
- `pnpm --filter @houston/host --filter @houston/domain test` — 175 + 17 files pass.
- `pnpm check` clean.

## Before / after

Before: a pool turn has no exec tool (`HOUSTON_CODE_EXECUTION=disabled`), and any workspace file a turn writes is dropped at sync-back.
After (with the cloud manifest change): a pool turn runs bash inside its own single-use pod; files it produces land in the object store and re-hydrate on the next turn.

Rollout requires the paired cloud change; with only this PR deployed, behavior is unchanged (`disabled` stays `disabled`).


cravenceiling:
## Review round 1 applied

Addressed four adversarial reviews (2× gpt-5.6-sol, 2× fable, high). Fixes in `285dd42` + `a8e6a9e`:

- **Env scrub** (`pool-bash.ts`): pool bash runs with an allowlisted child env (mirrors `claude-env.ts`), so a turn can't read the worker token / store secrets.
- **Unclaimed-turn rejection**: a single-use worker 400s an unclaimed real `/turn` (it would have run bash without spending the pod).
- **Admission straddle closed**: re-check `isDraining` after acquiring the slot; `spend` excludes shadow warm-ups.
- Marker requires `HOUSTON_HOME` + boot write-probe; single-use shutdown gets a force-exit deadline; sync uses `holdDeletesOnFailure` + surfaces skipped/conflicts; local-without-single-use warns+degrades instead of crash-looping; widened writes classify through the canonical `agentFileEventType`; draining `/health` returns 503 for readiness.

**Residual (tracked, see cloud #315 comment):** the on-disk spent marker is defeatable by a malicious turn (`rm` + nonzero exit); the robust fix is control-plane-authoritative spent detection (dispatch record). Env is scrubbed so the marker is the only in-pod trust item; the `restartCount>=3` backstop + fresh-VM-per-restart bound the exposure. Proposing CP-side detection as a fast-follow.

PR #1411, opened 2026-08-27
fix(host): local cron fires defer to the control-plane routine delivery (PRODUCT-1549)

cravenceiling:
Fixes PRODUCT-1549 (routine works at :20/:40, fails every :00 with `no_provider`).

## Root cause

On managed-cloud pods **two schedulers race every scheduled routine instant**, and which one wins decides which credentials the run gets:

- The **control plane's fire delivery** (`/agents/:id/routine-fires`) carries the routine creator's minted acting identity, so the turn resolves the **creator's own** credentials.
- The **pod-local cron scan** cannot mint identity (by design, HOU-976 D10), so its fire runs on the **shared team** credential scope.

The per-instant dedup burn arbitrates the race. At the top of the hour the control plane's delivery lags a few seconds behind the herd of due hourly routines, the pod-local scan wins the burn, and — when the agent's saved provider is connected on the creator's personal scope but not the team scope — the runtime correctly refuses to run (sticky saved pick, never silently switch) and the run is recorded `runtime 409 no_provider` in 0s. Verified in production logs for the reported routine: the failing :00 instants were pod-fired on the team scope (credential serve answered the team row, "org not connected"), while the succeeding :20/:40 instants were control-plane-delivered with the acting identity (personal rows served, turn ran). This also explains matching :00 failure bursts across many other workspaces.

## Fix

Local cron fires now wait out a 120s grace (`EXTERNAL_FIRE_GRACE_MS`) on deployments where the external fire delivery exists (managed-cloud pod-store marker; desktop and self-host are unchanged — their local cron is the only scheduler and still fires on due). The identity-carrying delivery therefore wins every live race, and the pod-local cron remains a genuine backstop that picks up any instant the delivery leaves unclaimed for 2 minutes. The scan's evaluation window is shifted whole, so consecutive ticks still tile exactly and the dedup key remains the true cron instant — an instant the delivery already burned is still deduped, never double-fired.

## Verification

Before (reproduces on a managed pod): a routine on a 20-minute schedule whose creator's provider is connected personally (not on the team scope) fails exactly the top-of-hour instants with `runtime 409 … no_provider` in 0s, while :20/:40 succeed.

After: the pod-local scan never fires an instant younger than 120s, so every live instant is fired by the control-plane delivery under the creator's identity; the :00 runs succeed. If the delivery is down, the pod fires the instant ~2 minutes late (backstop behavior, previously the racy path).

Tests: `scheduler-mode.test.ts` covers the grace deferral, the exactly-once fire after the grace, and dedup against an externally burned instant. Full `@houston/host` (1600), `@houston/runtime` (1398), `@houston/domain` (200) suites pass; Biome + typecheck clean.

### DIFF
--- a/packages/host/src/local/host.ts
+++ b/packages/host/src/local/host.ts
@@ -43,7 +43,11 @@ import { forward } from "../proxy/route";
 import { CredentialServeHealer } from "../routes/credential-healer";
 import { CUSTOM_OAUTH_CALLBACK_PATH } from "../routes/custom-integrations-oauth";
 import { ChannelRoutineFirer } from "../schedule/firer";
-import { type RoutineSchedulerMode, Scheduler } from "../schedule/scheduler";
+import {
+  EXTERNAL_FIRE_GRACE_MS,
+  type RoutineSchedulerMode,
+  Scheduler,
+} from "../schedule/scheduler";
 import { type ControlPlaneDeps, createControlPlaneServer } from "../server";
 import { syncSharedEndpoint } from "../shared-endpoint/sync";
 import { LocalWorkspaceStore } from "../store/local";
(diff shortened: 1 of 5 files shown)


cravenceiling:
Follow-up commit: the shifted evaluation window could look back up to one grace before process start, so a pod restarting within 120s after a delivered fire would have re-fired that instant (the fire lock is process-local). The left edge now clamps to scheduler start — a fresh replica never replays history, same as before the grace. The shift also moved from agent-scan into the scheduler tick, which owns startedAt.

Pool-path compatibility (checked, no change needed): the pool routine runner handles only fires whose agent is asleep (`Handled=false` falls back to the standing-pod delivery this PR fixes), an asleep agent runs no local cron, and the pool dispatcher's credential fetch already forwards the creator's acting token — so pool-served routines resolve personal credentials correctly and never race this path.

PR #1449, opened 2026-08-31
feat(agents): copy agent and Agent Store publish from settings, settings as default tab

cravenceiling:
## What

Three changes to the agent configuration screen:

1. **Publish to the Agent Store** row in the agent's Settings section. It reopens the existing share wizard (pick, anonymize, review, listing, share / manage when already listed), which had been dormant since the teams-first shell removed its last caller. The pick and review steps now spell out the privacy contract: only the picked items (job description, skills, routines, learnings) are shared, and conversations, files and connected accounts always stay private. The row only renders when identity is configured, since publishing needs a signed-in store account.

2. **Copy agent** row in the same section. The dialog asks for a name and a team (all of the user's teams are offered, including the agent's current one) and creates the duplicate through existing surfaces on both backends: agent-scoped portable export, in-browser unpack, create-with-seeds, then the shared team-move action files the copy in the chosen team. The copy carries the job description, skills, routines, learnings and color, plus the provider/model pin in the background; conversations and files are never part of the portable surface, so they are never copied.

3. **Settings is now the default section** when opening an agent's configuration. Job description was the previous landing tab; deep links keep naming their own section.

## Edge cases

- Agent names are unique per **workspace**, not per team, so copying next to the original always needs a new name. The dialog opens on the first free `<name> copy` / `<name> copy N`, validates live against the loaded agent list (taken, too long, invalid characters), and keeps the 409 race (a sibling takes the name mid-flight) as an expected-state toast.
- Copy failures keep the dialog open for a retry; the create is revealed optimistically (same contract as create/import) so a warming pod does not blank the flow.
- On the local backend the team placement is the layout write the move action already performs; server-backed it is the `PUT /agents/:slug/team` mutation with its rollback and expected-error handling.

## Tests

- New unit tests for the copy model (name suggestion incl. case-insensitivity, length cap, exhaustion; full-selection mapping) and for the new default-section rule.
- e2e: the settings drill-in now asserts landing on the manage section, and a new spec covers the copy dialog (pre-filled free name, inline conflict on a taken name, disabled create).
- Full gates green: biome, boundaries, parity, app typecheck + 3850 unit tests, locale sync (en/es/pt), web typecheck + shim parity, and the affected Playwright suites (agent-policy, team-view, team-manager-gate, agent-teams).


cravenceiling:
Follow-up pushed: the Settings tab now sits first in the rail (left of Job description) and carries the drilled h1. The copy dialog's taken-name error now explains the constraint: agent names are unique across the whole workspace (folder-derived ids), not per team, so a copy always needs its own name even in a different team.

cravenceiling:
Third commit: every agent row in the rail now carries an always-visible "..." menu in the row's affordance slot, after the count — [name] (7) ... — with Copy agent, Publish to the Agent Store (hidden when identity isn't configured) and Delete agent. Each opens the same surface the Settings section uses: the copy dialog, the share wizard, and the delete confirmation. ui/layout exposes the slot as data (SidebarItem.affordance); the library owns no menu, and the collapsed rail's hover flyout deliberately strips it. Board-home visual baselines re-recorded for darwin and linux (official Playwright container).

PR #1450, opened 2026-09-01
fix(ci): bake staging store site URL into staging web builds

cravenceiling:
Staging web builds (preview + the staging DMG flavor) bake `VITE_AGENTSTORE_GATEWAY_URL` to the staging gateway but leave `VITE_AGENTSTORE_SITE_URL` unset, so `STORE_SITE_URL` (`packages/web/src/engine-adapter/store-gateway.ts`) falls back to the prod catalog site. Browse/"view listing" links from a staging build then open the prod site, where staging-published agents 404.

The staging store site exists and serves the staging gateway's catalog, so this bakes its hostname into `web-staging.yml` and the `release.yml` staging flavor (prod flavor pinned to the same value as the code fallback, so behavior there is unchanged). Public hostnames, so literals — no new secrets.

Verified against staging: an authenticated publish returns `shareUrl` on the staging site host (composed server-side), the agent resolves on the staging catalog API and 404s on prod, and delete cleans it up.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -485,6 +485,9 @@ jobs:
       # steps aside and publish rides the engine base, while store-catalog reads
       # this value for browse.
       VITE_AGENTSTORE_GATEWAY_URL: ${{ matrix.flavor == 'staging' && secrets.HOSTED_ENGINE_URL_STAGING || secrets.HOSTED_ENGINE_URL }}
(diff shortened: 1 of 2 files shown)


cravenceiling:
Added the same default to `scripts/dev/staging.sh` — `pnpm dev:staging` had the identical fallback-to-prod-site gap (overridable via `VITE_AGENTSTORE_SITE_URL` in `.env.staging`).

PR #1471, opened 2026-09-02
fix(providers): serve Bedrock only ids a plain API key can invoke (PRODUCT-1641)

cravenceiling:
## Summary

Linear: PRODUCT-1641

A Bedrock user switching models in the chat picker hit two failure cards on a valid key. The host served pi-ai's full `amazon-bedrock` list (118 ids), and the picker folds a model's variants down to the shortest id, which for Bedrock is exactly the one Bedrock refuses:

- Bare `anthropic.claude-*` foundation ids: Bedrock serves Claude 4.x+ only through inference profiles, so these 400 with "on-demand throughput isn't supported" (the default-model half of this was PRODUCT-1477; the picker still offered them).
- Off-region inference profiles (`au.` / `eu.` / `jp.`, e.g. `au.anthropic.claude-opus-5`, the lexically-first survivor of the dedupe for Opus 5): 400 "The provided model identifier is invalid", which classified `unknown` (report-bug card + a Sentry error per attempt).

## Fix

- **Host catalog** (`packages/host/src/providers/bedrock-catalog.ts`, applied in `buildProviderCatalog`): `amazon-bedrock` now serves only `global.` cross-region profiles and Amazon's own `amazon.` Nova models. `us.` profiles are dropped too: keeping them would let the picker's shortest-id dedupe prefer `us.` over the `global.` twin and hide the curated/default `global.anthropic.claude-sonnet-4-6` row, and `global.` works from every endpoint region.
- **Classifier** (`packages/runtime/src/ai/provider-error.ts`): "provided model identifier is invalid" is now `model_unavailable`, and Bedrock gets `global.anthropic.claude-sonnet-4-6` as the suggested fallback (both for this body and the on-demand-throughput one), so the user sees the switch-model card with a one-click target.

Not done: a connect-time probe of every kept id. The kept set is the curated default (`global.anthropic.claude-sonnet-4-6`), Nova, and the other `global.` profiles.

## Tests

- `bedrock-catalog.test.ts`: pure filter, the two failing ids from the report are absent from the served catalog, every curated host Bedrock id + default survives.
- `provider-error.test.ts`: verbatim "provided model identifier is invalid" → `model_unavailable` with the global fallback; the existing on-demand-throughput case now asserts the fallback; the fallback never self-suggests.
- Ran: `pnpm check`, host + runtime vitest suites, `app` `tsgo --noEmit`, `provider-overrides-drift.test.ts`.

## Verify

Before: connect Bedrock with a plain API key, open the chat model picker, pick Claude Sonnet 4.6 (row id `anthropic.claude-sonnet-4-6`) or Claude Opus 5 (row id `au.anthropic.claude-opus-5`), send a message. Both fail; Opus 5 shows the generic report-bug card.

After: `GET /v1/catalog` lists only `global.*` and `amazon.*` ids under `amazon-bedrock`; the picker's Claude rows are the `global.` profiles and each completes a turn. A persisted off-region pick (e.g. an agent still on `au.anthropic.claude-opus-5`) now shows the switch-model card offering Claude Sonnet 4.6.

### DIFF
--- a/packages/host/src/providers/bedrock-catalog.test.ts
+++ b/packages/host/src/providers/bedrock-catalog.test.ts
@@ -1,0 +1,95 @@
+import type { Api, Model } from "@earendil-works/pi-ai";
+import { expect, test } from "vitest";
+import { PROVIDERS } from "../providers";
+import {
+  BEDROCK_PROVIDER_ID,
+  invokableBedrockModels,
+  isInvokableBedrockModelId,
+} from "./bedrock-catalog";
+import { buildProviderCatalog } from "./pi-catalog";
+
+function bedrockModel(id: string): Model<Api> {
+  return {
+    id,
+    name: id,
+    api: "bedrock-converse-stream",
+    provider: BEDROCK_PROVIDER_ID,
+    baseUrl: "",
+    reasoning: false,
+    input: ["text"],
+    cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0 },
+    contextWindow: 200_000,
+    maxTokens: 8_192,
+  } as Model<Api>;
+}
+
+test("keeps global inference profiles and Amazon's own models, in order", () => {
+  const kept = invokableBedrockModels([
+    bedrockModel("amazon.nova-pro-v1:0"),
+    bedrockModel("anthropic.claude-sonnet-4-6"),
+    bedrockModel("global.anthropic.claude-sonnet-4-6"),
+    bedrockModel("au.anthropic.claude-opus-5"),
+    bedrockModel("global.openai.gpt-5.6-sol"),
+    bedrockModel("us.anthropic.claude-opus-5"),
+  ]).map((m) => m.id);
+  expect(kept).toEqual([
+    "amazon.nova-pro-v1:0",
+    "global.anthropic.claude-sonnet-4-6",
+    "global.openai.gpt-5.6-sol",
+  ]);
+});
+
+test("bare Claude foundation ids and regional profiles are not invokable", () => {
+  // Bedrock 400s both on a valid key: the bare id with "on-demand throughput
+  // isn't supported", the off-region profile with "The provided model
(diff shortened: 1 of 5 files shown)


cravenceiling:
Live probe of the kept set against the reporting user's Bedrock API key from his pod (us-east-1, `POST /model/<id>/converse`, maxTokens 1), 2026-09-02:

| id | result |
|---|---|
| `global.anthropic.claude-sonnet-4-6` | 200 |
| `global.anthropic.claude-opus-4-8` | 200 |
| `global.anthropic.claude-sonnet-5` | 200 |
| `global.anthropic.claude-opus-5` | 200 |
| `global.anthropic.claude-fable-5` | 400 `data retention mode 'default' is not available for this model` |
| `us.anthropic.claude-sonnet-5` / `us.anthropic.claude-opus-5` | 200 |
| `eu.anthropic.claude-opus-5` | 400 `The provided model identifier is invalid.` |
| `amazon.nova-lite-v1:0` | 200 |
| `amazon.nova-premier-v1:0` | 400 on-demand throughput not supported |

So the `global.` Claude set this PR keeps is verified on a real key, with two follow-ups worth folding in here:

1. **`data retention mode`** is a third rejection a valid key hits (Fable 5 needs an AWS-side opt-in). It currently classifies `unknown`. Suggest adding it to `MODEL_UNAVAILABLE_PATTERNS` next to the new `provided model identifier is invalid` entry, with a test on the verbatim body above. The switch-model card with the global Sonnet 4.6 fallback is the honest one; the key is fine.
2. **`amazon.nova-premier-v1:0`** survives the `amazon.` prefix rule but is inference-profile-only. Harmless now that it lands on the fallback card, but worth an explicit drop or a `us.amazon.nova-premier-v1:0` swap if you want the Nova rows all to work.

Dropping `us.` to keep the fold on `global.` is the right call; both families served identically on this key.

PR #1508, opened 2026-09-07
fix(runtime): move the Codex default off gpt-5.5 and retire gpt-5.4 (PRODUCT-1695)

cravenceiling:
Closes PRODUCT-1695.

## Problem

Picking GPT-5.5 or GPT-5.4 on a ChatGPT (Codex) connection fails every turn with the "Model not available" card. Verified live on 2026-09-07: `chatgpt.com/backend-api/codex/responses` answers `gpt-5.5` with **404 "The model gpt-5.5 does not exist or you do not have access to it"** (over SSE and WebSocket, with pi-ai 0.85.1, with a raw request carrying Codex's own originator, and from the Codex CLI itself, which exhausts its 5 retries on the same 404) and `gpt-5.4` with **400 "not supported when using Codex with a ChatGPT account"**. `gpt-5.4` is gone from OpenAI's own model list; `gpt-5.5` is still listed, so it stays offered.

Houston's Codex default was `gpt-5.5` (runtime, domain, frontend), so a fresh ChatGPT connection failed its first turn. Codex CLI's built-in default is now `gpt-6-astra`.

## Change

- Codex default → `gpt-5.6-terra` in `config.codexModel`, the domain `DEFAULT_MODEL` and `PROVIDER_OVERRIDES.openai.defaultModel` (twins, comments cross-reference).
- `gpt-5.4` removed from `VISIBLE_MODELS.openai` and the domain `VALID_MODELS`: hidden from the picker, a stored pick migrates to the default with a diagnostic, an already-pinned conversation keeps working on the wire.
- Runtime classifier: a Codex `model_unavailable` carries `gpt-5.6-terra` as `suggested_fallback` (never the failing model itself), so the card offers a one-click switch instead of only "Pick another model".
- Chat panel's unsupported-model retry uses `getDefaultModel("openai")` instead of a hardcoded `gpt-5.5`.
- Copy: GPT-5.5 is "Previous-generation model." (en/es/pt), matching OpenAI's own description.

## Tests

- domain: retired `gpt-5.4` migrates to the default with a diagnostic; missing provider/model land on `gpt-5.6-terra`.
- runtime: Codex 404 and retired-model 400 both classify as `model_unavailable` with the `gpt-5.6-terra` fallback; fallback is null when the failing model is the fallback; providers without a known fallback still offer none.
- app: `gpt-5.4` is no longer a valid picker id; default is `gpt-5.6-terra`.

`pnpm check`, `pnpm --filter @houston/runtime --filter @houston/domain --filter @houston/host test`, `cd app && pnpm test && pnpm tsgo --noEmit && pnpm check-locales`, `pnpm --filter houston-web typecheck` all pass.

### DIFF
--- a/app/src/components/use-agent-chat-panel.tsx
+++ b/app/src/components/use-agent-chat-panel.tsx
@@ -2132,7 +2132,7 @@ export function useAgentChatPanel({
             }}
             onSwitchModel={
               isModelUnsupported
-                ? () => selectModel("openai", "gpt-5.5")
+                ? () => selectModel("openai", getDefaultModel("openai"))
                 : undefined
             }
           />
--- a/app/src/lib/provider-overrides.ts
+++ b/app/src/lib/provider-overrides.ts
@@ -203,6 +203,9 @@ export const VISIBLE_MODELS: Readonly<Record<string, ReadonlySet<string>>> = {
     "gpt-5.4",
     "gpt-5.4-mini",
   ]),
+  // No gpt-5.4: OpenAI retired it for ChatGPT accounts (400 "not supported
+  // when using Codex with a ChatGPT account"); a conversation already pinned
+  // to it keeps working on the wire, it just can't be picked anymore.
   openai: new Set([
     "gpt-6-astra",
     "gpt-5.5",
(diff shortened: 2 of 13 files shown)


cravenceiling:
Closing: the rejection is OpenAI-side (chatgpt.com answers gpt-5.5 with 404 and gpt-5.4 with 400 for ChatGPT accounts, the Codex CLI fails identically), and Houston already runs the picked model and shows the error card. Changing the default or hiding models is not the fix.

PR #1601, opened 2026-09-16
fix(host): a routine the agent creates runs on the chat's provider, never the runtime's last-used one

cravenceiling:
Fixes [PRODUCT-1849](https://linear.app/houston-ai/issue/PRODUCT-1849).

## Problem

daniel@gethouston.ai created "Resumen de correos nuevos" from a chat running on Sonnet 5. The routine screen showed **Local model · Gemma4:12b · Sin conectar**: the routine would fire on a local model the user never connected.

- `save_routine` has no provider parameter, so every agent-created routine is saved unpinned.
- An unpinned routine does not run on the agent's configured model. The runtime resolves it from its own `settings.json` `activeProvider`, the last provider any chat on that runtime used (`resolveModel`, `packages/runtime/src/ai/providers.ts`).
- The composer auth-gates its initial provider; nothing on the routine path did.

## Fix

The host already records each turn it starts (`routes/live-turn.ts`) for mission ancestry, plan mode and acting identity. It now also records the **provider/model/effort pin the send asked for** (`turnPinOf`, read off the shared body memo, so the engine still receives the same bytes). The agent's `POST /sandbox/routines/save` create path pins a new routine to that pair when the agent named none: a routine born in a chat runs on what that chat runs on, which is by construction a provider the user connected.

- Explicit pins from the agent are kept.
- Updates never re-pin: an existing pin or a deliberate "follows the agent" is the user's choice.
- A save with no recorded turn, or a send with no pin, stays unpinned as before.
- Stored in pi's canonical dialect (`openai` → `openai-codex`), the one the fire path resolves.
- Programmatic fires (routine, trigger, mission first turn) record their `TurnPin` too, so a mission's first turn that saves a routine inherits its pair.

## Not covered

Existing unpinned routines keep following the runtime's last-used provider. The routine screen shows the resolved pair honestly (PRODUCT-1475), so a user can pin one from the Model row.

## Tests

- `routes/routines-sandbox.test.ts` (new): inherits the chat pair, canonical dialect, explicit pin kept, no pin / no live turn stays unpinned, update never re-pins.
- `routes/turn-body.test.ts` (new): `turnPinOf`.
- `routes/live-turn.test.ts`: pin rides the record, `liveTurnPin` from a `TurnPin`.
- `pnpm check` exit 0; `@houston/host` 259 files / 2546 tests pass.


cravenceiling:
Closing at Daniel's request.

PR #1603, opened 2026-09-16
feat(store): a listing carries the agent's routines, and an installed one says what it still needs

cravenceiling:
## Why

The export wizard's pick step offers an agent's routines and its copy promises they are shared, but `irFromPortable` never put them in the AgentIR, the contract had no field for them, and the gateway dropped the key. A published agent silently lost every routine, and an installer got none. (Screenshot in the thread: "Say 5 hourly" and the Gmail webhook routine both toggled on and both lost.)

## What

**Contract** (`packages/agentstore-contract`): AgentIR gains an additive optional `routines` array (max 32, unique ids). `irVersion` stays `2.0.0` because zod strips unknown keys and `encoding/json` ignores them, so shipped readers keep installing everything else, whereas bumping the literal would make them reject new listings. Each routine is `{ id, name, prompt, wake, chatMode?, suppressWhenSilent? }` with `wake` discriminated on `kind`: `schedule` (cron), `composio` (lowercase toolkit, uppercase triggerSlug, triggerConfig object at most 8000 chars serialized), `webhook` (no fields). Normalize derives ids and fixes slug case, and unions composio toolkits into `integrations`. Secret scan covers routine name, prompt and triggerConfig. JSON schema and fixture updated.

**Domain** (`store-ir.ts` + new `store-ir-routines.ts`): publish maps a Houston routine to the IR and drops enabled, provider/model/effort, connected account, webhook key prefix, setup chat and creator. Install maps back (enabled, defaults, fresh timestamps) and runs `normalizeRoutines`, so the import wizard, id re-mint and seed write all work unchanged.

**App**: an installed Composio routine whose app is not connected on this account shows Not connected + Connect right away in the routine header chip instead of spinning to "couldn't confirm". Gated on the viewer being the routine's creator, same scope rule as `useRoutineProviderHealth`, so a teammate's routine is never accused. Webhook routines already land unminted with Create webhook address. The imported agent's setup mission prompt now lists each routine that still needs an app connected or an address created, so the agent asks for it in its first message. Existing locale keys only.

**Store site**: `AgentDetailScreen` (ui/store) gains a Routines section; the agent page shows each routine's name and a plain wake label; the receiving-assistant install instructions list them. Inventory v86 + CHANGELOG, `check:parity` green.

## Depends on

gethouston/cloud#372 (gateway accepts and serves `routines`) must deploy first. Until then routines are dropped by the gateway exactly as today.

## Checks

`pnpm check`, contract / domain / host / runtime vitest, `pnpm typecheck`, app `tsgo` + 4491 node tests + `check-locales`, houston-web typecheck + tests, agentstore typecheck + tests, `check:parity`, `check:boundaries`: all green.


cravenceiling:
Closed at Daniel's request.

